PR #3, opened 2022-12-12
Start with telemetry popup

anastasiya1155:
When being initialized Sentry makes an API request to the Sentry server which is unwanted before the user gives their consent. Solution: Before showing the onboarding flow show the Telemetry popup and ask the user for their consent; when the user changes permissions close the current Sentry client or reinitialize it depending on the choice. Also, pass the release version to Sentry when initializing it.
Minor bug fix: copying text to the clipboard doesn't work if the user doesn't click on a button, so this function call is removed now and shouldn't throw an error in the console.


ipauler (approved):

ggordonhall:
@anastasiya1155 Do we need to accompany this with any changes to the Tauri logic?

anastasiya1155:
> @anastasiya1155 Do we need to accompany this with any changes to the Tauri logic?

I don't think so, looks like BE does not make this initial request to sentry

PR #4, opened 2022-12-12
UI fixes and improvements

anastasiya1155:
A list of changes:
- Input now doesn't make a sound when submitting a query for the first time
- Hovering over files in repo view is now consistent
- Removed filters from repo view (not useful atm)
- Render markdown for readme
- Add a back button to all pages
- Fixed an issue with multiple open sidebars in repo view after opening a couple of files
- Go back to the path search after opening a file should now be fixed
- Show telemetry popup before onboarding flow


ipauler (dismissed):
LGTM

PR #9, opened 2022-12-12
Add histogram-based logging for long-running files

rsdy:
Add behind the `debug` feature flag an option to log severe outliers during analytics, including file size.

This is useful, because it seems like large files are hogging 1 thread, and can take up to 40s to complete (e.g. Kubernetes has a lot of generated `.go` files of several MiB in size). Identifying outliers will be helpful in addressing the issue.

If this seems useful, we may want to convert this into a default feature, as the runtime impact should be fairly lightweight.

### DIFF
--- a/server/bleep/Cargo.toml
+++ b/server/bleep/Cargo.toml
@@ -25,35 +25,26 @@ name = "queries"
 harness = false
 
 [dependencies]
+
+# core
 tantivy = { version = "0.18.1", features = ["mmap"] }
 tokio = { version = "1.21.2", features = ["macros", "process", "rt", "rt-multi-thread", "io-std", "io-util", "sync", "fs"] }
+futures = "0.3.25"
 rayon = "1.6.0"
 clap = { version = "4.0.26", features = ["derive"] }
-tower-http = {version = "0.3.4", features = ["cors", "catch-panic"]}
 tracing = "0.1.37"
 tracing-subscriber = { version = "0.3.16", features = ["env-filter", "registry"] }
 color-eyre = "0.6.2"
 
-# for debugging tokio
+# for debugging
 console-subscriber = { version = "0.1.8", optional = true }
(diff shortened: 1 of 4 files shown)


nerdypepper (approved):
until we can improve scope-queries to operate faster on larger files (which i do think we can), a short term solution would be to prevent calculating intelligence on files that take too long to parse (which probably means they contain tons of tokens; e.g.: generated and/or minified).

server/bleep/Cargo.toml
```diff
@@ -89,6 +87,7 @@ replace_with = "0.1.7"
 relative-path = "1.7.2"
 git2 = "0.15.0"
 dunce = "1.0.3"
+histogram = { version = "0.6.9", optional = true }
```
ggordonhall, line 90: Can you add a comment explaining that this is used for logging when the debug feature is enabled?

server/bleep/src/indexes/file.rs
```diff
@@ -330,126 +347,153 @@ impl Indexer<File> {
...
+            .match_indices('\n')
+            .flat_map(|(i, _)| u32::to_le_bytes(i as u32))
+            .collect::<Vec<_>>();
+
+        let lines_avg = buffer.len() as f64 / buffer.lines().count() as f64;
+        let last_commit = repo_info.last_commit_unix_secs;
+
+        trace!("writing document");
+
+        let buf_size = buffer.len();
```
ggordonhall, line 459:
```suggestion
        #[cfg(feature = "debug")]
        let buf_size = buffer.len();
```

server/bleep/Cargo.toml
rsdy: `optional` dependencies are usually picked up by `features`, like here: https://github.com/BloopAI/bloop/pull/9/files#diff-f8919bb455c366daf0d703faef73f63e745e0a06dde47f6e62b5eb88c78687b1R10

PR #11, opened 2022-12-13
Check Segment write key env var

anastasiya1155:
When building an app for production we must ensure that 'Segment write key' is present in the environment, otherwise, analytics will not work. With this change, FE will throw an error during build process if this env is not found.


ipauler (dismissed):

ggordonhall (approved):

stunningpixels:
Can we include ENV for Sentry on this PR please

anastasiya1155:
> Can we include ENV for Sentry on this PR please

We don't have env for Sentry. The [dsn for Sentry is public info](https://docs.sentry.io/product/sentry-basics/dsn-explainer/#:~:text=DSNs%20are%20safe%20to%20keep%20public) and there is no reason to put it to env, prod/dev flag in events is set from the env of the app

ggordonhall:
@anastasiya1155 discuss this and as implemented we'll receive prod Sentry errors if users modify the source code and build bloop from source. We don't want this, so we should move to a model that we only initialise Sentry if a DSN is set as an environment variable. Our DSN can be set in CI.

This is require some changes. I think we'll need to conditionally initialise Sentry (otherwise will error with invalid or empty DSN)

anastasiya1155:
> @anastasiya1155 discussed this and as implemented we'll receive prod Sentry errors if users modify the source code and build bloop from source. We don't want this, so we should move to a model that we only initialise Sentry if a DSN is set as an environment variable. Our DSN can be set in CI.
> 
> This is require some changes. I think we'll need to conditionally initialise Sentry (otherwise will error with invalid or empty DSN)

Added the fix for FE, should I do it for Rust as well, or are you already on it?

anastasiya1155:
> Can you tackle? Could make an `initialize_sentry` Tauri command.

Can we have some env var for Rust and just use it to initialize Sentry instead of rewriting the functionality?

apps/desktop/.env.example
```diff
@@ -1 +1,2 @@
-VITE_SEGMENT_WRITE_KEY=
+VITE_SEGMENT_WRITE_KEY_DEV=
+VITE_SEGMENT_WRITE_KEY_PROD=
```
ggordonhall, line 2: Can you add `SENTRY` env vars here?

apps/desktop/src-tauri/src/main.rs
```diff
@@ -99,12 +79,33 @@ async fn main() {
...
+            initialize_sentry
         ])
         .run(tauri::generate_context!())
         .expect("error running tauri application");
 }
 
+#[tauri::command]
+fn initialize_sentry(dsn: String, environment: String) {
+    if let Some(_) = sentry::Hub::current().client() {
+        return;
```
ggordonhall, line 92:
```suggestion
        tracing::info!("Sentry has already been initialized");
        return;
```

stunningpixels (approved):

PR #14, opened 2022-12-13
implement regex-filtering at the collector level

nerdypepper:


server/bleep/src/webserver/query.rs
```diff
@@ -85,112 +86,32 @@ impl ApiQuery {
...
+            } else if FileReader.query_matches(q) {
+                return FileReader
+                    .execute(&indexes.file, &queries, &self)
+                    .await
+                    .map_err(|e| EndpointError::internal(e.to_string().into()));
+            } else if OpenReader.query_matches(q) {
+                return OpenReader
+                    .execute(&indexes.file, &queries, &self)
+                    .await
+                    .map_err(|e| EndpointError::internal(e.to_string().into()));
```
calyptobai, line 109: Looks like this change breaks queries with multiple different types, e.g. `foobar or symbol:foo`. Can we revert to the old behaviour here?

server/bleep/src/webserver/query.rs
```diff
@@ -395,16 +410,32 @@ impl ExecuteQuery for FileReader {
...
+            .await?;
+
+        let filter_regexes = queries
+            .iter()
+            .filter(|q| self.query_matches(q))
             .filter_map(|q| Some(q.path.as_ref()?.regex().expect("failed to parse regex")))
             .collect::<SmallVec<[_; 2]>>();
 
-        Ok(results
+        let data = results
```
ggordonhall, line 438: I may be missing something, but doesn't it also make sense to filter regexes at the Collector level for `FileReader` and `RepoReader`. Don't they suffer the same problem that `top_docs` will return documents that contain all trigrams in the query, but don't necessarily match the regex, leading to pagination issues for `path:foo` and `repo:bar` type searches.

server/bleep/src/indexes/file.rs
```diff
@@ -96,9 +98,11 @@ impl File {
         let repo_disk_path = builder.add_text_field("repo_disk_path", STRING);
         let repo_ref = builder.add_text_field("repo_ref", STRING | STORED);
         let repo_name = builder.add_text_field("repo_name", trigram.clone());
+        let raw_repo_name = builder.add_bytes_field("raw_repo_name", FAST);
         let relative_path = builder.add_text_field("relative_path", trigram.clone());
 
         let content = builder.add_text_field("content", trigram.clone());
+        let raw_content = builder.add_bytes_field("raw_content", FAST);
```
calyptobai, line 105: This has the effect of putting *all* file data in a fast field, essentially defeating the point of a fast field. Is it possible to do this without marking `raw_content` as `FAST`, or does the collector strategy require this?

server/bleep/src/webserver/query.rs
calyptobai: Makes sense. This has the side effect of interpreting `repo:foo or bar` as a repo-only query. Can you add a comment above this condition section for future clarity as to what's happening?

server/bleep/src/indexes/file.rs
rsdy: Collectors can only work with fast fields. It's not inherently wrong, I think if we can't measure perf regression here, let's roll with this. However, might be a good idea to try and avoid having both `content` and `raw_content`.

server/bleep/src/webserver/query.rs
```diff
@@ -244,6 +165,39 @@ pub(super) struct QueryMetadata {
     total_count: Option<usize>,
 }
 
+#[derive(Default, Serialize, Deserialize, Debug)]
+pub struct ResultStats {
```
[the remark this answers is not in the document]
rsdy: I'd rather see `QueryMetadata` as `PagingMetadata` we can potentially reuse for other pageable things in the query, eg. when `/repos/` gets too large.

rsdy (commented):
I like this, but definitely needs more testing. If you look at `FileReader` and `RepoReader` side-by-side, they're _very_ similar, though, so maybe a few interfaces can be still cleaned up to cut down on noise.

server/bleep/src/webserver/query.rs
rsdy: I'd rather see `QueryMetadata` as `PagingMetadata` we can potentially reuse for other pageable things in the query, eg. when `/repos/` gets too large.

nerdypepper: @ggordonhall 
>I may be missing something, but doesn't it also make sense to filter regexes at the Collector level for FileReader and RepoReader

yes, you are right, i have added the bytes field for this but somehow forgotted about the filter collector.

nerdypepper: @rsdy 
>I'd rather see QueryMetadata as PagingMetadata we can potentially reuse for other pageable things in the webserver, eg. when /repos/ gets too large.

yep, it makes sense for that to be common across most of our endpoints. i believe this is in response to @ggordonhall's other comment:
>I think it makes sense to merge this into the QueryMetadata struct (langs and repos become fields). This will require a change to the client

i will keep these units separate for now; in the form of `ResultStats` and `PagingMetadata`.

server/bleep/src/indexes/file.rs
nerdypepper: yes, unfortunately collectors can only access fast fields/facet fields (which might be worth exploring for our `lang` and `repo_name` fields).

another option i have considered is reconstructing the relevant portions of the document from the term indices (+ term dictionary). with our terms here being trigrams, i am not sure if this makes much sense, but we may be able to retain the speed guarantees and avoid the extra field.

ggordonhall (approved):

nerdypepper:
@rsdy 
>I like this, but definitely needs more testing. If you look at FileReader and RepoReader side-by-side, they're very similar, though, so maybe a few interfaces can be still cleaned up to cut down on noise.

sorry, missed this, but i think @calyptobai intends to clean up the `ExecuteQuery` and `DocumentReader` interfaces very soon, maybe some of the duplication will go away with that PR.

nerdypepper:
@rsdy @calyptobai need your opinion on the index-size issue. using additional fast-fields blows up index size considerably (~2x on larger repos). IMO, we can improve on that by tweaking the filter collector, introduced by this PR, down the line.

PR #16, opened 2022-12-13
Add new `profiler` profile for better performance work

rsdy:
More aggressive optimization for the `release` profile, while adding a new one to help with profiling in depth.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,10 +7,10 @@ members = [
 [profile.release]
 debug = false
 strip = "debuginfo"
-panic = "abort"
 lto = true
 
 [profile.profiler]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -5,4 +5,12 @@ members = [
 ]
 
 [profile.release]
+debug = false
+strip = "debuginfo"
+panic = "abort"
```
calyptobai, line 10: This is quite harsh, it means that our app will completely die on a panic that could be caught by the webserver, and e.g. reported to sentry.
```suggestion
```

calyptobai (commented):
Nice addition, left 2 comments

PR #24, opened 2022-12-14
Fix spurious flashing issue in windows

nerdypepper:
Windows runs child processes with `cmd.exe`, which simply _has_ to create a new window. Luckily, we have a flag: `CREATE_NO_WINDOW` to prevent that.

### DIFF
--- a/server/bleep/src/remotes.rs
+++ b/server/bleep/src/remotes.rs
@@ -1,4 +1,6 @@
-use std::path::Path;
+#[cfg(windows)]
+use std::os::windows::process::CommandExt;
+use std::{path::Path, process::Command};
 
 use anyhow::{bail, Context, Result};
 use dashmap::mapref::one::Ref;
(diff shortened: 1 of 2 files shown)


server/bleep/src/remotes.rs
```diff
@@ -73,11 +75,20 @@ pub(in crate::remotes) fn git_pull(
         Some(git2::build::CheckoutBuilder::new().force()),
     )?;
 
-    match tokio::process::Command::new("git")
```
nerdypepper, line 76: @rsdy do we need `tokio::process::Command` here?

ggordonhall (approved):

yqs112358:
It is strange that on my PC the problem still exists. I have installed Windows Terminal and the terminal window may flash over sometimes.
I can confirm this issue because the terminal window will always flash when cloning a new repo. It is stable reproducible.

My PC is Windows 11 23H2 with Windows Terminal 1.17.11461.0

PR #25, opened 2022-12-14
Fix duplicate repo creation during re-indexing

calyptobai:


ggordonhall (approved):
This makes sense and addresses the issue. Do we also need this step for the `File` index and, if not, why?

rsdy (approved):

rsdy:
@ggordonhall file uses a cache to identify which files have been updated since the last indexing, and will rotate those based on file paths.

Technically, we could also *skip indexing* a repo based on some other heuristic instead of deleting & recreating it every time. For instance, if the `Repository` object in the pool has a `last_index_time > 0`, just move on. This approach would create fewer Tantivy writes, but I don't believe in the grand scheme of things the difference would be material.

PR #27, opened 2022-12-14
[RFC] Interruptable scope_graph generation

rsdy:
Reduce the parsing time to 0.1s. For some reason the previous 1s limit causes weird runaway loops parsing highly complex files. We might want to escalate this to upstream with a test suite.

# Verifying results

[983d36c](https://github.com/BloopAI/bloop/pull/27/commits/983d36c3caf965396bbcfb4a7077988c2dac9604) reduces the parsing timeout to 0.1s on paper. This somehow avoids choking the indexing process, and extreme outliers for the entire worker function. Ultimately it cuts total processing from 7m40s to 3m20s for the testset.

### DIFF
--- a/server/bleep/src/intelligence.rs
+++ b/server/bleep/src/intelligence.rs
@@ -66,7 +66,7 @@ impl<'a> TreeSitterFile<'a> {
     }
 
     /// Produce a lexical scope-graph for this TreeSitterFile.
-    pub fn scope_graph(self) -> Result<ScopeGraph, TreeSitterFileError> {
+    pub async fn scope_graph(self) -> Result<ScopeGraph, TreeSitterFileError> {
         let query = self
             .language
             .scope_query
@@ -74,6 +74,8 @@ impl<'a> TreeSitterFile<'a> {
             .map_err(TreeSitterFileError::QueryError)?;
         let root_node = self.tree.root_node();
 
-        Ok(ResolutionMethod::Generic.build_scope(query, root_node, self.src, self.language))
+        Ok(ResolutionMethod::Generic
+            .build_scope(query, root_node, self.src, self.language)
+            .await)
     }
 }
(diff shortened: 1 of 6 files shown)


rsdy:
Profiling the bottlenecked parts that are constrained to a single thread, using `perf`, shows the following:

```
47.86%  bleep            bleep              [.] ts_tree_cursor_parent_node        
  41.14%  bleep            bleep              [.] ts_tree_cursor_current_status     
  10.60%  bleep            bleep              [.] ts_language_symbol_metadata       
   0.06%  bleep            bleep              [.] ts_query_cursor__advance    
```

Trying to repro on macOS with additional debug info, the call stack is not proving super helpful:
Screenshot 2022-12-14 at 21 44 50

There seems to be a tight loop somewhere we can't interrupt.

calyptobai:
Interesting approach. I'm curious if we can fix the performance issue in an alternate way with the use of a new data structure? If there's no obvious fix, and if this does show better performance in tests, it seems like a decent stopgap.

rsdy:
[983d36c](https://github.com/BloopAI/bloop/pull/27/commits/983d36c3caf965396bbcfb4a7077988c2dac9604) reduces the parsing timeout to 0.1s on paper. This somehow avoids choking the indexing process, and extreme outliers for the entire `worker` function.

Unclear if we need the rest of the hacks for safety.

server/bleep/src/intelligence.rs
```diff
@@ -52,7 +52,7 @@ impl<'a> TreeSitterFile<'a> {
             .map_err(|_| TreeSitterFileError::LanguageMismatch)?;
 
         // do not permit files that take >1s to parse
```
calyptobai, line 54:
```suggestion
        // do not permit files that take >100ms to parse
```

nerdypepper:
>Interesting approach. I'm curious if we can fix the performance issue in an alternate way with the use of a new data structure?

The issue we are currently seeing is fundamental to tree-sitter from what I understand. Executing the tree-sitter-query DSL program (every `scopes.scm` file) simply takes too much time on complex ASTs (`ts_tree_cursor_parent_node` indicates tree-traversal).

server/bleep/src/intelligence.rs
nerdypepper, line 54:
```suggestion
        // do not permit files that take >100ms to parse
```

nerdypepper (approved):

nerdypepper (dismissed):

PR #30, opened 2022-12-15
Refactor app navigation

ipauler:
## Changes:
Refactor results page:
- All pages now are managed by `/search` page since they all depend on search results
- File modal view is moved out of navigation logic and used only in results view
- Add full page view which is now used in Repository view for viewing files
- Add lang filters on repo view page

Refactor navigation:
- Replace browser navigation now all search results are managed by `/search` page
- Add custom virtual navigation with persistent store


client/src/hooks/useAppNavigation.tsx
```diff
@@ -0,0 +1,142 @@
...
+  );
+
+  const buildQuery = (navigationItem: NavigationItem) => {
+    const { query, path, repo, type } = navigationItem;
+    switch (type) {
+      case 'repo':
+      case 'full-result':
+        return `open:true repo:${repo} ${path ? `path:${path}` : ''}`;
+      case 'search':
+        return query;
```
anastasiya1155, line 69: I think we should encodeURIComponent here to fix issues with having special symbols in the url

client/src/pages/Results/index.tsx
```diff
@@ -155,94 +92,78 @@ const ResultsPage = () => {
...
-      </>
-    );
-  };
+  }, [fileResultData]);
+
+  useEffect(() => {
+    // if (!isFileNav) {
+    //   ref.current?.scrollTo({ top: 0, left: 0, behavior: 'smooth' });
+    // }
+  }, [page]);
```
anastasiya1155, line 126: this useEffect should either be not empty or removed

client/src/pages/Skeleton/index.tsx
```diff
@@ -67,7 +67,7 @@ export const ResultsPreviewSkeleton = () => {
 };
 
 const mainContainerStyle = { height: 'calc(100vh - 8rem)' };
-const Skeleton = () => {
+const Index = () => {
```
anastasiya1155, line 70: I believe this renaming was accidental

client/src/pages/Results/index.tsx
```diff
@@ -86,66 +68,21 @@ const ResultsPage = () => {
...
-
-    setInitialLoad(false);
-  }, [data]);
 
-  useEffect(() => {
-    if (!isFileNav) {
-      ref.current?.scrollTo({ top: 0, left: 0, behavior: 'smooth' });
+  const onResultClick = useCallback((repo: string, path?: string) => {
+    if (path) {
+      fileModalSearchQuery(`open:true repo:${repo} path:${path}`);
```
anastasiya1155, line 74: shouldn't we also get hoverable ranges for this file?

stunningpixels:
Nothing unexpected when testing the app. One thing, I think there's a 'flash' when navigating repo view:

- Go to main screen
- Click on repo
- Click on folder
- You should experience a quick flash before content loads

Is this due to the loader showing and then content loading after? Would be good to make this smoother somehow

ipauler:
@stunningpixels yes i've seen that flash and it's related to loader 
will fix that

client/src/pages/Results/index.tsx
ipauler: fixed

client/src/pages/Skeleton/index.tsx
ipauler: fixed

client/src/pages/Results/index.tsx
ipauler: fixed

PR #33, opened 2022-12-16
Don't report indexing as "done" before Tantivy commit is over

rsdy:
Additionally, deleting logic a bit more robust here, as it doesn't neeed to be aware of multiple repositories in this function.


ggordonhall (approved):
LGTM

PR #34, opened 2022-12-19
omit hidden files from ctags calculations

nerdypepper:
files such as `.eslint.rc` and `.gitignore` were being omitted from the index.

### DIFF
--- a/server/bleep/src/ctags.rs
+++ b/server/bleep/src/ctags.rs
@@ -40,7 +40,6 @@ fn find_files(path: &Path) -> Vec<String> {
     WalkBuilder::new(path)
         .ignore(true)
         .git_ignore(true)
-        .hidden(false)
         .build()
         .filter_map(|res| match res {
             Ok(de) => Some(de),


ggordonhall:
We use a similar walker [here](https://github.com/BloopAI/bloop/blob/f6fce1d251985880c6258ae8f3dcf27f13541fb5/server/bleep/src/ctags.rs#L40) and [here](https://github.com/BloopAI/bloop/blob/f6fce1d251985880c6258ae8f3dcf27f13541fb5/server/bleep/src/language.rs#L22). Can you think of situations where we'd want to calculate language and symbols for hidden files?

nerdypepper:
>We use a similar walker [here](https://github.com/BloopAI/bloop/blob/f6fce1d251985880c6258ae8f3dcf27f13541fb5/server/bleep/src/ctags.rs#L40) and [here](https://github.com/BloopAI/bloop/blob/f6fce1d251985880c6258ae8f3dcf27f13541fb5/server/bleep/src/language.rs#L22). Can you think of situations where we'd want to calculate language and symbols for hidden files?

good catch, i can see language being useful (`.eslintrc.json` can be detected as json), we can skip hidden files for ctags calculations.

rsdy:
What's blocking the merge of this?

nerdypepper:
my negligence. ill enable hidden files for `get_language_info` and merge this in.

nerdypepper:
so this changeset is already on `main`, i propose that we disable ctags for hidden files however; updating this PR in accordance.

PR #38, opened 2022-12-20
Semantic search

rsdy:


.github/workflows/client-test.yml
```diff
@@ -3,7 +3,7 @@ name: Client Tests
 on:
   pull_request:
     types: [opened, synchronize]
-    branches: [main]
+    branches: [main, semantic]
```
nerdypepper, line 6: we can revert this change with the merge of `semantic`. cc @anastasiya1155.

.github/workflows/server-test.yml
```diff
@@ -2,7 +2,7 @@ name: Server Unit Tests
 
 on:
   pull_request:
-    branches: [main]
+    branches: [main, semantic]
```
nerdypepper, line 5: likewise, this hunk can be reverted.

.github/workflows/tauri-test.yml
```diff
@@ -3,7 +3,7 @@ name: Tauri Tests
 on:
   pull_request:
     types: [opened, synchronize]
-    branches: [main, dev]
+    branches: [main, semantic]
```
nerdypepper, line 6: revert this hunk on merge.

apps/desktop/src-tauri/src/qdrant.rs
```diff
@@ -0,0 +1,256 @@
...
+    path::Path,
+    process::{Child, Command},
+    time::Duration,
+};
+
+use tauri::{plugin::Plugin, Runtime};
+
+use super::relative_command_path;
+
+pub(super) const QDRANT_CONFIG: &str = r#"
```
nerdypepper, line 12: store this as a yaml file elsewhere and use `include_str!` instead.

apps/desktop/src-tauri/src/qdrant.rs
```diff
@@ -0,0 +1,256 @@
...
+                .replace(
+                    "{{ STORAGE }}",
+                    &qdrant_dir.join("storage").to_string_lossy(),
+                )
+                .replace(
+                    "{{ SNAPSHOTS }}",
+                    &qdrant_dir.join("snapshots").to_string_lossy(),
+                ),
+        )
+        .unwrap();
```
nerdypepper, line 177: there seems to be some trivial templating going on here, we can avoid using string replacements here by doing two things:

- put the template config into `config.yaml` and use rust's format string syntax, i.e., use `{storage}` for template holes
- achieve templating with the `format!` and `include_str!` macros, `format!(inculude_str!("./config.yaml"), storage = storage, snapshots = snapshots)`

server/bleep/src/ctags.rs
```diff
@@ -51,7 +51,7 @@ fn find_files(path: &Path) -> Vec<String> {
         })
         .filter(|de| matches!(de.file_type(), Some(ft) if ft.is_file()))
         .map(|de| {
-            dunce::canonicalize(de.path())
+            crate::canonicalize(de.path())
```
nerdypepper, line 54: @llogiq i am not sure if there is much to be gained here by conditionally compiling dunce as a dependency. internally, it uses a cfg flag to point to `fs::canonicalize` on non-windows targets (IIRC), and other targets should end up with zero extra compilation. for what its worth, dunce is a 100 line library, we might as well vendor it in if required.

server/bleep/src/intelligence/language.rs
```diff
@@ -55,6 +55,9 @@ pub struct TSLanguageConfig {
     /// Compiled tree-sitter scope query for this language.
     pub scope_query: MemoizedQuery,
 
+    /// Compiled tree-sitter chunk query for this language.
+    pub chunk_query: Option<MemoizedQuery>,
+
```
nerdypepper, line 60: this can be removed.

nerdypepper (changes requested):

server/bleep/src/ctags.rs
llogiq: I know it's a small crate, but adding it unconditionally means we download and compile it everywhere. Yes, even macros are compiled. So even if it's just a second to download it and another half to compile, we end up winning one and a half second per non-windows build. A small win, yes, but those tend to add up.

nerdypepper: makes sense.

anastasiya1155 (dismissed):

apps/desktop/src-tauri/src/qdrant.rs
nerdypepper: https://github.com/BloopAI/bloop/pull/96

ipauler (approved):

PR #52, opened 2023-01-04
Better recovery after shutdown

rsdy:
Previously shutdowns would occasionally cause invalid JSON structures for the state files. In this PR we fix this by making these writes atomic from the FS perspective by first writing to a new file, then moving it into place.

Also, since we dump the raw memory structure into the state JSONs, we need to reset the `sync_status` on repos that were being indexed when the shutdown happened, and queue them again.

### DIFF
--- a/server/bleep/src/state.rs
+++ b/server/bleep/src/state.rs
@@ -6,6 +6,7 @@ use crate::{
 };
 use clap::Args;
 use dashmap::DashMap;
+use rand::Rng;
 use relative_path::RelativePath;
 use secrecy::{ExposeSecret, SecretString};
 use serde::{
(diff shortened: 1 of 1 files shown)


server/bleep/src/state.rs
```diff
@@ -515,9 +518,14 @@ impl StateSource {
...
 }
 
-pub fn pretty_write_file<T: Serialize + ?Sized>(path: &Path, val: &T) -> Result<(), RepoError> {
-    let file = std::fs::File::create(path)?;
+pub fn pretty_write_file<T: Serialize + ?Sized>(
+    path: impl AsRef<Path>,
+    val: &T,
+) -> Result<(), RepoError> {
+    let tmpfile = path.as_ref().with_extension("new");
+    let file = std::fs::File::create(&tmpfile)?;
```
llogiq, line 526: We should use `File::options().create_new().open(tmpfile)` instead to avoid overwriting an existing file. If we want to be really fancy, we could use a counter or random extension with a retry mechanism.

llogiq:
Looks good in general, there's only a small security-related suggestion.

ggordonhall (approved):

llogiq (approved):

PR #57, opened 2023-01-05
Remove redundant norm calculation

ggordonhall:
Removes a redundant norm calculation, and the `ndarray_linalg` crate that is depends on.


rsdy (approved):
yay!

PR #59, opened 2023-01-05
chunk by number of tokens instead of lines

llogiq:
This tries to maximize the number of tokens per chunk while still splitting at newlines. Overlap is not yet implemented and lines that are longer than the maximum number of chunks will still silently be put into one single chunk (and likely truncated by the model). Next step will be splitting overlong lines by white space and then implementing overlap.

### DIFF
--- a/server/bleep/src/semantic/chunk.rs
+++ b/server/bleep/src/semantic/chunk.rs
@@ -86,7 +86,10 @@ pub fn tree_sitter<'a, 'l>(src: &'a str, lang_id: &'l str) -> Result<Vec<Chunk<'
 }
 
 fn point(src: &str, byte: usize) -> Point {
-    let line = src.as_bytes()[..byte].iter().filter(|&&b| b == b'\n').count();
+    let line = src.as_bytes()[..byte]
+        .iter()
+        .filter(|&&b| b == b'\n')
+        .count() + 1; // we count from line 1
     let column = if let Some(last_nl) = src[..byte].rfind('\n') {
         byte - last_nl
     } else {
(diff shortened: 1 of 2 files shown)


server/bleep/src/semantic/chunk.rs
```diff
@@ -76,6 +85,62 @@ pub fn tree_sitter<'a, 'l>(src: &'a str, lang_id: &'l str) -> Result<Vec<Chunk<'
...
+    let mut first_newline = 0;
+    let mut last_newline = 0;
+    let mut result = Vec::new();
+    for newline in (0..(offset_len - 1)).filter(|&i| src[offsets[i].1 .. offsets[i + 1].0].contains('\n')) {
+        if newline - first_newline > max_tokens {
+            add_chunks(&mut result, src, offsets, first_newline, last_newline, max_tokens);
+            first_newline = (last_newline + 1).min(offset_len);
+        }
+        last_newline = newline;
+    }
```
nerdypepper, line 137: could you help me understand what is going on here? specifically, how does this check: `newline - first_newline > max_tokens` work?

llogiq: Sure! So our chunks have a size limit: max_tokens. We want to fit as many lines as we can into the chunk. So we have to catch the first occasion where we have more than `max_tokens` tokens, go back one line (that's why we keep the `last_newline` around) and chunk until then. If we added a chunk we reset our `first_newline` to the position after the lines we just put into the chunk.

llogiq:
Note that this code still has a problem: Huggingface's tokenizers return char positions instead of byte positions. If we have non-ascii characters, this will likely crash because we may index into the midst of a codepoint. We can either switch to aleph-alpha-tokenizers or reconstruct the byte positions from the char positions.

rsdy (approved):
Is there a reason you don't want to bring in the aleph alpha tokenizer in this PR? I'm assuming the interface should be similar.

server/bleep/src/semantic/chunk.rs
nerdypepper: understood, makes sense now.

nerdypepper (approved):

ggordonhall (approved):

llogiq:
aleph-alpha-tokenizers was built for their specific use case, and I haven't touched it for quite some time. I don't want to break something in other parts of the code I haven't had a deep look at in my first week :grin:.

PR #61, opened 2023-01-06
Error when semantic search returns no snippets

ggordonhall:
Throw an error when semantic search returns no results, rather than passing an empty `Vec` to the OpenAI API. We should display a standard message on the client in this case.


anastasiya1155 (commented):
When can it happen? I tried removing cache and even with no index there are still 10 snippets

ggordonhall:
> When can it happen? I tried removing cache and even with no index there are still 10 snippets

That's a bug that I've encountered. Probably caused by a dirty Qdrant shutdown.

PR #62, opened 2023-01-09
Make search input wider

anastasiya1155:
Now search input will take more space in the header so that long queries fit into it better

### DIFF
--- a/client/src/components/NavBar/User/index.tsx
+++ b/client/src/components/NavBar/User/index.tsx
@@ -23,7 +23,7 @@ const NavBarUser = ({ shareFiles, isSkeleton }: Props) => {
   }, []);
 
   return (
-    <div className="flex flex-row flex-1">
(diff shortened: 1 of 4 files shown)


anastasiya1155:
@stunningpixels can you check it and confirm that it's better now?

ipauler (approved):

stunningpixels:
Screenshot 2023-01-10 at 09 51 04

Can we standardise max widths pls. Other than that, looks good to me.

anastasiya1155:
> Screenshot 2023-01-10 at 09 51 04
> 
> Can we standardise max widths pls. Other than that, looks good to me.

it's not as simple as it may seem. The main body has a left sidebar that can be collapsed when it shows filters, while the navbar has icons on the left and on the right, which makes input positioning a bit hard. I will have to spend a bit more time thinking about the correct position.

stunningpixels:
Ok let's merge

PR #64, opened 2023-01-09
add `lang` and `repo` filters to NL search

nerdypepper:
this enables the `lang` and `repo` filters for natural language search. note that this currently does not work with regex filters such as `repo:/blo.*/`, if a regex filter is encountered, it is ignored and no filter is applied.

the NL search query grammar is very lenient; a query of the form:

```
lang:tsx background repo:bloop color
```` 
is interpreted as:

```
background color (lang:tsx) (repo:bloop)
```
instead of raising a syntax error.

### DIFF
--- a/server/bleep/src/query/parser.rs
+++ b/server/bleep/src/query/parser.rs
@@ -22,6 +22,27 @@ pub enum Target<'a> {
     Content(Literal<'a>),
 }
 
+#[derive(Default, Clone, Debug, PartialEq, Eq)]
+pub struct NLQuery<'a> {
+    pub repo: Option<Literal<'a>>,
+    pub lang: Option<Cow<'a, str>>,
+    pub target: Option<Literal<'a>>,
+}
+
+impl<'a> NLQuery<'a> {
+    pub fn repo(&self) -> Option<&Cow<'_, str>> {
+        self.repo.as_ref().and_then(|t| t.as_plain())
+    }
+
+    pub fn lang(&self) -> Option<&Cow<'_, str>> {
+        self.lang.as_ref()
+    }
+
+    pub fn target(&self) -> Option<&Cow<'_, str>> {
+        self.target.as_ref().and_then(|t| t.as_plain())
+    }
+}
+
 impl<'a> Query<'a> {
     /// Merge this query with another, overwriting current terms by terms in the new query, if they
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: server/bleep/src/query/parser.rs [hunk 1]
```diff
@@ -22,6 +22,27 @@ pub enum Target<'a> {
...
+#[derive(Default, Clone, Debug, PartialEq, Eq)]
+pub struct NLQuery<'a> {
+    pub repo: Option<Literal<'a>>,
+    pub lang: Option<Cow<'a, str>>,
+    pub target: Option<Literal<'a>>,
+}
+
+impl<'a> NLQuery<'a> {
+    pub fn repo(&self) -> Option<&Cow<'_, str>> {
+        self.repo.as_ref().and_then(|t| t.as_plain())
```
nerdypepper, line 34: if `repo` is a regex here, it is silently converted to a `None` instead of raising a parse error.

server/bleep/src/semantic.rs
```diff
@@ -121,7 +124,39 @@ impl Semantic {
...
+                    match_value: MatchValue::Text(value).into(),
+                }),
+                ..Default::default()
+            }
+            .into()
+        };
+
+        let repo_filter = nl_query
+            .repo()
+            .map(|r| make_kv_filter("repo".to_string(), r.to_string()));
```
nerdypepper, line 149:
```suggestion
            .map(|r| make_kv_filter("repo_name".to_string(), r.to_string()));
```

server/bleep/src/query/parser.rs
ggordonhall: Will this term just be ignored then? E.g. `How do we parse a query? repo:/bloop/` would just be executed as `How do we parse a query?`

ggordonhall (approved):
Lgtm 👍

rsdy (approved):

server/bleep/src/query/parser.rs
nerdypepper: yes, this term will just be ignored altogether. i haven't figured out regex filters yet, unsure if its possible through qdrant directly either, what we might need is some kind of query-builder that translates an NLQuery down to a Qdrant query in a better way.

PR #65, opened 2023-01-09
Add e2e tests

ipauler:
Add playwright e2e tests


anastasiya1155 (approved):

stunningpixels:
Would be very cool if we could isolate the testing of features of the app from onboarding

ipauler:
@stunningpixels sorry what do you mean ?
right now we have separate onboarding tests and separate functionality tests

PR #68, opened 2023-01-09
Chunk by tokens

llogiq:
This time, we won't try to split lines into zero and all the other tokens. :stuck_out_tongue:

I also added some assertions, which appear to fail for my tests, unfortunately.

### DIFF
--- a/server/bleep/src/semantic.rs
+++ b/server/bleep/src/semantic.rs
@@ -19,7 +19,7 @@ use qdrant_client::{
     },
 };
 use rayon::prelude::*;
-use tracing::{debug, trace, info};
+use tracing::{debug, info, trace};
 
 pub mod chunk;
 
@@ -205,7 +205,7 @@ impl Semantic {
                     len = chunk.len(),
                     big_chunk = chunk.len() > 800,
                     "new chunk",
-                );nfo!("embedding {} tokens {:?}", length, chunk
+                );
 
                 match self.embed(chunk.data) {
                     Ok(ok) => Some(PointStruct {
--- a/server/bleep/src/semantic.rs
+++ b/server/bleep/src/semantic.rs
@@ -146,7 +146,8 @@ impl Semantic {
         buffer: &str,
         lang_str: &str,
     ) {
-        let chunks = chunk::trivial(buffer, 15); // line-wise chunking, 15 lines per chunk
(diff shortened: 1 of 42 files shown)


llogiq:
Now this looks like it should be working. In my tests chunking the bloop repo, I get *9984* chunks with the trivial by-line chunking without overlap and *8266* chunks by tokens with 50% overlap. However, we may still have problems, I get a lot of warnings from onnxruntime.

I may want to refactor the whole thing a bit to pull the embedding to the front of the process so we can reuse it after chunking. Otherwise I would appreciate some help debugging this.

server/bleep/src/semantic.rs
```diff
@@ -186,9 +189,17 @@ impl Semantic {
...
+            repo_name,
+            relative_path,
+            buffer,
+            &*self.tokenizer,
+            self.config.embedding_input_size,
+            15,
+            self.config.overlap.unwrap_or(chunk::OverlapStrategy::Partial(0.5)),
+        );
+        let repo_plus_file = repo_name.to_owned() + "\t" + relative_path + "\n";
+        info!(chunk_count = chunks.len(), "found chunks");
```
ggordonhall, line 202:
```suggestion
        debug!(chunk_count = chunks.len(), "found chunks");
```

server/bleep/src/semantic.rs
```diff
@@ -199,7 +210,7 @@ impl Semantic {
                     "new chunk",
```
ggordonhall, line 210: Do we still need this `debug` statement?

server/bleep/src/semantic/chunk.rs
```diff
@@ -108,3 +378,49 @@ pub fn trivial(src: &str, size: usize) -> Vec<Chunk<'_>> {
...
+                &file.path().to_string_lossy(),
+                &src,
+                &tokenizer,
+                max_tokens,
+                max_lines,
+                OverlapStrategy::Partial(0.5),
+            )
+            .len();
+            let linewise = trivial(&src, 15).len();
+            dbg!((path, tokenwise, linewise));
```
ggordonhall, line 420: Can we remove these `dbg!` statements?

ggordonhall (approved):
Lgtm. Just a couple of nits

server/bleep/src/semantic.rs
llogiq: No. Removing it.

rsdy (approved):

PR #70, opened 2023-01-10
Rework answer-api to be 2-staged

nerdypepper:


nerdypepper:
The sister PR on `bloop/answer-api` is: https://github.com/BloopAI/answer-api/pull/8

server/bleep/src/webserver/answer.rs
```diff
@@ -94,34 +98,141 @@ pub async fn handle(
...
-            snippets: snippets.clone(),
-        })
-        .send()
+    let answer_api_client = AnswerAPIClient::new(
+        &format!("{}/q", app.config.answer_api_base),
+        &target,
+        &params.user_id,
+    );
+
+    let res1 = answer_api_client
```
ggordonhall, line 107: Can we have a more descriptive name for this? relevant_snippet_index`?

server/bleep/src/webserver/answer.rs
```diff
@@ -94,34 +98,141 @@ pub async fn handle(
...
+        .stage_1(&snippets)
+        .await
+        .map_err(|e| super::error(ErrorKind::Internal, e.to_string()))?
+        .index;
+
+    let processed_snippet = &snippets[res1];
+
+    let res2 = answer_api_client
```
ggordonhall, line 115: Similarly can call this `answer`

server/bleep/src/webserver/answer.rs
```diff
@@ -94,34 +98,141 @@ pub async fn handle(
...
+}
+
+#[derive(serde::Deserialize)]
+struct Stage2Response {
+    answer: String,
+}
+
+const DELIMITER: &str = "######";
+impl AnswerAPIClient {
+    async fn stage_1(&self, snippets: &[api::Snippet]) -> Result<Stage1Response, reqwest::Error> {
```
ggordonhall, line 184: Let's call this `select_snippet` (and rename the corresponding `Response` struct)

server/bleep/src/webserver/answer.rs
```diff
@@ -94,34 +98,141 @@ pub async fn handle(
...
+            })
+            .collect::<String>();
+
+        prompt += &format!(
+            "\nAbove are {} code snippets separated by \"{DELIMITER}\". \
+            Your job is to answer which snippet index best answers the question. Your\
+            reply must obey the JSON schema {{ \"index\": number}}.
+
+            Q: What icon do we use to clear search history?
+            A: {{ \"index\": 0 }}
```
ggordonhall, line 202: This'll need to be tested but we don't need the model to return JSON at all. The instruction could be updated to `Your reply should be a single number` and the example could be `A: 0` directly.

server/bleep/src/webserver/answer.rs
```diff
@@ -94,34 +98,141 @@ pub async fn handle(
...
+            #####
+
+            Above is a code snippet.\
+            Your job is to answer the questions about the snippet,\
+            giving detailed explanations. Cite any code used to\
+            answer the question formatted in GitHub markdown and state the file path.\
+            Your reply must obey the JSON schema {{ \"answer\": string }}
+
+            Q: What icon do we use to clear search history?
+            A: {{ \"answer\": \"We use the left-double chevron icon.\" }}
```
ggordonhall, line 229: Same here, there's no need for the model to generate JSON.

server/bleep/src/webserver/answer.rs
```diff
@@ -94,34 +98,141 @@ pub async fn handle(
...
+            Q: {}
+            A:",
+            snippets.len(),
+            self.query,
+        );
+
+        self.send(prompt).await?.json::<Stage1Response>().await
+    }
+
+    async fn stage_2(&self, snippet: &api::Snippet) -> Result<Stage2Response, reqwest::Error> {
```
ggordonhall, line 213: Rename `answer`

server/bleep/src/webserver/answer.rs
```diff
@@ -94,34 +98,141 @@ pub async fn handle(
...
-        )
-    })?;
-
-    let selected_snippet = snippets.remove(selection.data.index as usize);
+        .map_err(|e| super::error(ErrorKind::Internal, e.to_string()))?
+        .answer;
+
+    // reorder snippets
+    let selected_snippet = snippets.remove(res1);
     snippets.insert(0, selected_snippet);
```
llogiq, line 123: Is it important that the other snippets keep their order? If not, we could use `swap` here. Otherwise, I believe that `snippets[..(res1 + 1)].rotate_right(1)` should do the same thing with less overhead.

llogiq (approved):
Looks good, I only have a tiny nit.

server/bleep/src/webserver/answer.rs
nerdypepper: ah good point, i gave this a go but i was not able to guarantee the output format. i think supplying an example Q/A pair should help with that though.

server/bleep/src/webserver/answer.rs
nerdypepper: the order of the other snippets is not important, we can absolutely use `swap` here.

nerdypepper:
we no longer require it to respond in JSON, `bloop/answer-api` has been updated accordingly.

PR #71, opened 2023-01-10
Skip indexing large, vendored, or generated files

calyptobai:
Using some heuristics, generated and vendored files are now skipped for indexing. These are done per-extension, using heuristic regexes. To start, this contains a small set of common patterns. More languages can extend the vendored list to incorporate other patterns.


server/bleep/src/indexes/file.rs
```diff
@@ -153,6 +156,64 @@ impl File {
...
+            .collect()
+    });
+
+    match VENDOR_PATTERNS.get(&*ext) {
+        None => true,
+        Some(rxs) => !rxs.iter().any(|r| r.is_match(&path.to_string_lossy())),
+    }
+}
+
+// Empirically calculated using:
```
rsdy, line 211: What's the corpus you have done this on? 30C/line seems a bit low?

rsdy (approved):

server/bleep/src/indexes/file.rs
calyptobai: Tested using the `bleep` source code here, surprised me as well

PR #73, opened 2023-01-11
Add segment logging

ggordonhall:
Adds Segment. If configured, for each call to the `/answer` endpoint will track the query, the snippet deemed most relevant and the LLM response.

### DIFF
--- a/server/bleep/src/lib.rs
+++ b/server/bleep/src/lib.rs
@@ -268,7 +268,7 @@ impl Application {
         }
 
         // configure segment
-        let segment = config.segment_key.clone().map(Segment::new);
+        let segment = Arc::new(config.segment_key.clone().map(Segment::new));
 
         let config = Arc::new(config);
         let semantic = Semantic::new(&config.model_dir, &config.qdrant_url).await?;
--- a/server/bleep/src/lib.rs
+++ b/server/bleep/src/lib.rs
@@ -280,7 +280,7 @@ impl Application {
             background: BackgroundExecutor::start(config.clone()),
             semantic,
             config,
-            segment: Arc::new(segment),
+            segment,
         })
     }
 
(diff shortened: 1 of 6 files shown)


server/bleep/src/webserver/answer.rs
```diff
@@ -129,11 +134,36 @@ pub async fn handle(
...
+            .await
+            .expect("Could not send to Segment")
+    }
+
     Ok::<_, Json<super::Response<'static>>>(Json(super::Response::Answer(AnswerResponse {
         snippets,
         selection: api::Response {
             data: api::DecodedResponse {
-                index: 0 as u32, // the relevant snippet is always placed at 0
+                index: 0_u32, // the relevant snippet is always placed at 0
```
rsdy, line 166:
```suggestion
                index: 0u32, // the relevant snippet is always placed at 0
```

server/bleep/src/lib.rs
```diff
@@ -247,6 +267,9 @@ impl Application {
                 .map_err(|existing| anyhow!("ctags binary already set: {existing:?}"))?;
         }
 
+        // configure segment
+        let segment = config.segment_key.clone().map(Segment::new);
```
rsdy, line 271:
```suggestion
        let segment = Arc::new(config.segment_key.clone().map(Segment::new));
```

server/bleep/src/lib.rs
```diff
@@ -257,6 +280,7 @@ impl Application {
             background: BackgroundExecutor::start(config.clone()),
             semantic,
             config,
+            segment: Arc::new(segment),
```
rsdy, line 283:
```suggestion
            segment,
```

server/bleep/src/webserver/answer.rs
```diff
@@ -129,11 +134,36 @@ pub async fn handle(
     // reorder snippets
     snippets.swap(relevant_snippet_index, 0);
 
+    if let Some(ref segment) = *app.segment {
```
rsdy, line 137: Can we move this logic into the `Segment` struct like `segment.send(segment::Track { ... }).await`, and the struct into a new file? Would be a bit cleaner, and contains all the relevant logic.

rsdy (changes requested):

rsdy (approved):

PR #83, opened 2023-01-13
reduce build time: Initial small wins

llogiq:
I also fixed the benchmarks en passant; I noticed they were broken for a while.

### DIFF
--- a/server/bleep/src/indexes/file.rs
+++ b/server/bleep/src/indexes/file.rs
@@ -176,7 +176,7 @@ impl Indexable for File {
             .open_walker()
             .filter_map(|entry| match entry {
                 Ok(de) => match de.file_type() {
-                    Some(ft) if ft.is_file() => Some(dunce::canonicalize(de.into_path()).unwrap()),
+                    Some(ft) if ft.is_file() => Some(crate::canonicalize(de.into_path()).unwrap()),
                     _ => None,
                 },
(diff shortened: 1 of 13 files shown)


ggordonhall (approved):

server/bleep/src/indexes/file.rs
```diff
@@ -167,7 +167,7 @@ impl Indexable for File {
         repo: &Repository,
         repo_info: &RepoHeadInfo,
         writer: &IndexWriter,
-    ) -> Result<()> {
+    ) -> Result<()> {
```
rsdy, line 170: Empty spaces at the end of line!

rsdy (approved):

server/bleep/src/indexes/file.rs
ggordonhall: We need to get the CI back up. Have to configure qdrant builds: https://github.com/BloopAI/bloop/actions/runs/3902002219/jobs/6664534799

llogiq: Oops, fixed.

llogiq:
rebased.

llogiq:
Didn't we have git-lfs before?

llogiq:
Now it appears we are throttled again, this time downloading onnxruntime. We should probably cache it on the CI image.

PR #84, opened 2023-01-13
remove user_id from answer-api requests

nerdypepper:
answer-api no longer uses segment, and therefore no longer recognizes the `user_id` request field.


ggordonhall (approved):
Good spot

PR #85, opened 2023-01-13
tune token count for completions

nerdypepper:
this helps the app produce some truly impressive completions. the performance takes a hit however, it could take upwards of 20s to produce an answer.

some example queries: 
- with an index of `bloop`: `auth flow frontend`
- with an index of `bloop`: `how are scope graphs built`

a sample response for the second query:

---

Scope graphs are used to represent lexical scopes and names in a single syntax tree.The function `ScopeGraph::new` is used to construct a scope graph given a `TextRange` and language index `lang_id`:

```rust 
pub fn new(range: TextRange, lang_id: usize) -> Self {
    let mut graph = Graph::new();
    let root_idx = graph.add_node(NodeKind::scope(range));
    Self {
        graph,
        root_idx,
        lang_id,
    }
}
```
This function creates an empty graph, inserts a root scope as a node, and annotates it with the given `range` and `lang_id`. 

The `insert_local_scope` and `insert_local_def` functions can be used to insert a local scope and definition respectively into the scope graph. 

`insert_local_scope` takes `LocalScope` and inserts it into the scope graph under the parent scope, or the scope that encompasses that `LocalScope`. 

`insert_local_def` takes `LocalDef` and inserts it into the scope graph under the scope where it is defined. 

`insert_hoisted_def` takes a `LocalDef` and can insert it into the parent scope of the defining scope, or the scope where the `LocalDef` is defined. 

`insert_ref` is used to insert a `Reference` into the scope graph. It will traverse all the scopes from the current scope to the root scope and find candidate definitions with the same name as the `Reference`. In addition, if both the `Reference` and definition have a `symbol_id`, it will only form an edge if the namespaces of the two match. 

Finally, `scope_by_range` and `parent_scope` are helper functions for finding scopes in the scope graph.
`scope_by_range` takes a `TextRange` and a `start` node index and will return the smallest scope encompassing that `TextRange`, starting from the `start` node specified. 
`parant_scope` takes a node index of a scope and will return the parent scope of that node, or `None` if there is none. 

server/bleep/src/intelligence/scope_resolution.rs

---

which is highly accurate if i say so myself.

### DIFF
--- a/server/bleep/src/webserver/answer.rs
+++ b/server/bleep/src/webserver/answer.rs
@@ -1,4 +1,5 @@
 use axum::{extract::Query, response::IntoResponse, Extension, Json};
+use reqwest::StatusCode;
 use utoipa::ToSchema;
 
 use crate::{indexes::reader::ContentDocument, query::parser, state::RepoRef, Application};
@@ -116,7 +117,13 @@ pub async fn handle(
     let relevant_snippet_index = answer_api_client
         .select_snippet(&snippets)
         .await
-        .map_err(super::internal_error)?
+        .map_err(|e| match e.status() {
+            Some(StatusCode::SERVICE_UNAVAILABLE) => super::error(
+                ErrorKind::UpstreamService,
+                "service is currently overloaded",
+            ),
+            _ => super::internal_error(e),
+        })?
         .text()
         .await
         .map_err(super::internal_error)?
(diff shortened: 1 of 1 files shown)


server/bleep/src/webserver/answer.rs
```diff
@@ -292,18 +299,28 @@ impl AnswerAPIClient {
...
     }
 }
+
+// openai claims 4 characters = 1 token for english,
+// but code typically clobbers the token count,
+// we get no more than 3 characters per token.
+//
+// TODO: this is a heuristic, a realistic tokenizer
+// can tune this better, rust_tokenizers::Gpt2Tokenizer?
+fn token_count_estimate(input: &str) -> u32 {
```
ggordonhall, line 324: Yes, the pretrained `GPT2Tokenizer` should be a better estimate. Could quickly run a side-by-side to see whether token estimates correlate: https://huggingface.co/docs/transformers/model_doc/openai-gpt

ggordonhall (approved):

server/bleep/src/webserver/answer.rs
nerdypepper: will leave this as a TODO for now, and create a new ticket to tackle this.

nerdypepper: ill extract the tokenizer config from gpt2, and load that into our app for token count predictions.

PR #88, opened 2023-01-13
Increase `open:` query performance on large repositories

calyptobai:
With the `tensorflow` repository, this takes `open:` query time down from ~1 second to ~20 milliseconds, a 50x speedup.

The central idea behind this change involves introducing dummy documents into the file index, representing directories. This allows us to search for folders, without having to infer their existence due to other documents.

Before, 1.028 seconds:
```
curl -v 'localhost:7878/q?q=open:true%20repo:tensorflow'  0.00s user 0.00s system 10% cpu 1.028 total
```

After, 19 milliseconds:
```
curl -v 'localhost:7878/q?q=open:true%20repo:tensorflow'  0.00s user 0.00s system 47% cpu 0.019 total
```

Additionally, a bug with open queries of the form `some/valid/file/with/trailing/slash.txt/` resulting in panics was fixed.

### DIFF
--- a/server/bleep/src/indexes/file.rs
+++ b/server/bleep/src/indexes/file.rs
@@ -1,7 +1,7 @@
 use std::{
     collections::{HashMap, HashSet},
     ops::Not,
-    path::{Path, PathBuf},
+    path::{Path, PathBuf, MAIN_SEPARATOR},
     sync::Arc,
 };
 
@@ -44,7 +44,7 @@ use crate::{
 };
 
 struct Workload<'a> {
-    file_disk_path: PathBuf,
+    entry_disk_path: PathBuf,
     repo_disk_path: &'a Path,
     repo_ref: String,
     repo_name: &'a str,
@@ -61,8 +61,8 @@ pub struct File {
     #[cfg(feature = "debug")]
     histogram: Arc<RwLock<Histogram>>,
 
-    // Path to the indexed file on disk
-    pub file_disk_path: Field,
+    // Path to the indexed file or directory on disk
+    pub entry_disk_path: Field,
     // Path to the root of the repo on disk
     pub repo_disk_path: Field,
     // Path to the file, relative to the repo root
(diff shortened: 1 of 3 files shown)


server/bleep/src/indexes/file.rs
```diff
@@ -246,28 +244,31 @@ impl Indexable for File {
...
-            };
-
-            debug!(?file_disk_path, "queueing file");
-            if let Err(err) = self.worker(workload, writer) {
-                warn!(%err, ?file_disk_path, "indexing failed; skipping");
-            }
-        });
+        walker
+            .into_par_iter()
+            .chain(Some(repo.disk_path.clone()))
```
nerdypepper, line 249: when walking some path `p`, does the walker not add `p` itself to the results?

server/bleep/src/webserver/query.rs
```diff
@@ -625,6 +620,41 @@ impl ExecuteQuery for OpenReader {
...
+        let collector = BytesFilterCollector::new(
+            indexer.source.raw_relative_path,
+            move |b| {
+                let Ok(relative_path) = std::str::from_utf8(b) else {
+                    return false;
+                };
+
+                // Check if *any* of the relative paths match. We can't compare repositories here
+                // because the `BytesFilterCollector` operates on one field. So we sort through this
+                // later. It's unlikely that a search will use more than one open query.
```
nerdypepper, line 640: this is achievable by chaining multiple filter collectors: `ByteFilterCollector::new(field1, filter1, BytesFilterCollector(field2, filter2, empty_collector))`

nerdypepper (dismissed):
this is working great, seeing some good speedups when testing on milvus. not sure if the repository sifting code is too important, but i've left a comment if so.

ggordonhall (dismissed):
Tested and this works really well. Can't reproduce that duplicate path error (must have been caused by a corrupted index) - this looks good to me 👍

server/bleep/src/indexes/file.rs
calyptobai: Good point, I misunderstood the behaviour. Fixed in [a4f1e18](https://github.com/BloopAI/bloop/commit/a4f1e1838538bd1675d81a51c85f9a6e846dc54b)

PR #90, opened 2023-01-16
use gpt-2 tokenizer to estimate token count accurately

nerdypepper:
using gpt2 to estimate the number of tokens we have left for completions should help us maximize the number of tokens we use, and avoid accidentally exceeding the token count.


server/bleep/src/webserver/answer.rs
```diff
@@ -146,7 +152,17 @@ pub async fn handle(
             .await
             .map_err(super::internal_error)?;
 
-        let grown_text = grow(&doc, relevant_snippet, 60);
+        let mut grow_size = 40;
```
nerdypepper, line 155: this change allows for us to grow the snippet dynamically until we reach 2000 tokens or 200 lines, which ever comes first.

the reason for setting the 200 line limit is to prevent small files (<100) lines from causing this to loop infinitely.

server/bleep/src/webserver/answer.rs
```diff
@@ -317,18 +337,18 @@ impl AnswerAPIClient {
...
-// openai claims 4 characters = 1 token for english,
-// but code typically clobbers the token count,
-// we get no more than 3 characters per token.
-//
-// TODO: this is a heuristic, a realistic tokenizer
-// can tune this better, rust_tokenizers::Gpt2Tokenizer?
-fn token_count_estimate(input: &str) -> u32 {
-    input.len().saturating_div(3) as u32
+        // do not let the completion cross 2500 tokens
+        let max_tokens = max_tokens.clamp(1, 500);
```
nerdypepper, line 350: regardless of the amount of wiggle room we have for completions, we set the limit to 500 here. this should hopefully keep the response times down; though in practice, the response times seem to be independent of the completion token count, and dependent on the extent of crowding of the API.

ggordonhall (approved):

PR #91, opened 2023-01-16
Track each step of the query pipeline in Segment

ggordonhall:
This PR updates our Segment query logging. Now we track each step of the query pipeline as a separate Segment property:
- User query
- Select prompt (includes all snippets returned by semantic search)
- Most relevant snippet
- Explain prompt
- Answer

### DIFF
--- a/server/bleep/src/webserver/answer.rs
+++ b/server/bleep/src/webserver/answer.rs
@@ -4,14 +4,14 @@ use tracing::{error, info};
 use utoipa::ToSchema;
 
 use crate::{
-    indexes::reader::ContentDocument, query::parser, semantic::Semantic, state::RepoRef,
-    Application,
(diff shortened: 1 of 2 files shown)


server/bleep/src/webserver/answer.rs
```diff
@@ -337,7 +335,15 @@ impl<'a> AnswerAPIClient<'a> {
             A:",
             snippet.relative_path, snippet.text, self.query
         );
-        let tokens_used = self.semantic.gpt2_token_count(&prompt);
+        prompt
+    }
+
+    async fn select_snippet(&self, prompt: &str) -> Result<reqwest::Response, reqwest::Error> {
+        self.send(prompt.to_string(), 1).await
```
nerdypepper, line 342: just a thought, `AnswerAPIClient::send` should accept a `&str` and the call site needn't pass owned data, not important to put in this PR.

nerdypepper (approved):

PR #92, opened 2023-01-17
Find on page shortcut

anastasiya1155:
When pressing cmd/cntr+f a search input will appear that allows users to search for some text on the page, much like the behavior in the browser. Pressing Esc should hide the input and remove all highlights.


ipauler (dismissed):

anastasiya1155:
@stunningpixels waiting for your review

PR #93, opened 2023-01-17
remove overlapping snippets from qdrant's response

nerdypepper:
before sending snippets for selection, we trim the snippet collection by removing overlapping snippets and limiting the number of snippets to 2 per file. 

the algorithm used to remove overlapping snippets is the [activity-selection](https://en.wikipedia.org/wiki/Activity_selection_problem) algorithm. here, the line-intervals of each snippet are analogous to the time-intervals of each activity. this algorithm helps us maximize the number of snippets selected per file.

### DIFF
--- a/server/bleep/src/webserver/answer.rs
+++ b/server/bleep/src/webserver/answer.rs
@@ -1,3 +1,5 @@
+use std::collections::HashMap;
+
 use axum::{extract::Query, response::IntoResponse, Extension, Json};
 use reqwest::StatusCode;
 use tracing::{error, info};
@@ -30,6 +32,7 @@ pub mod api {
         pub end_line: usize,
         pub start_byte: usize,
         pub end_byte: usize,
+        pub score: f32,
     }
 
     #[derive(Debug, serde::Serialize, serde::Deserialize)]
@@ -64,6 +67,8 @@ pub struct AnswerResponse {
     pub selection: api::Response,
 }
 
+const SNIPPET_COUNT: usize = 14;
+
 pub async fn handle(
     Query(params): Query<Params>,
     Extension(app): Extension<Application>,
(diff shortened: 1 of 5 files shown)


llogiq:
I'd suggest making the maximum number of snippets dependent on file size.

E.g. in the Rust compiler, there are files with 10k lines and others with just over a 100.

Perhaps a logarithmic calculation or sqrt(loc) might make sense?

nerdypepper:
>I'd suggest making the maximum number of snippets dependent on file size.

thats a good point, but instead of normalizing depending on file size, maybe we should normalize the count of snippets per file (in the search result list)? for now, ill implement some kind of weighting based on file size.

stunningpixels:
I think there's a bug, if I use a lang: filter I only get a handful of results

nerdypepper:
we are no longer guaranteed to return exactly `params.limit` number of results, since we do some post processing after the query step. this is similar to the pagination issue we had earlier:

- query for 10 results from the index
- the post-process step omits overlapping results, limits to N results per file etc.
- we end up showing <=10 results to the user

handling this gracefully requires some thought, we need to implement this at the query level. some stop-gap solutions (for a `limit` of 10):

- query for more than 10 results, say 20, post-process, limit the output to 10
- repeatedly re-query until we end up with 10 results

server/bleep/src/webserver/answer.rs
```diff
@@ -89,28 +95,99 @@ pub async fn handle(
...
+    }
+
+    // limit the number of snippets per file to atmost 20% of the total results
+    let per_file_limit = crate::div_ceil(4 * SNIPPET_COUNT, 5);
+    tracing::debug!(%per_file_limit, "setting per-file limit");
+    let mut snippets = snippets_by_file
+        .into_iter()
+        .inspect(|(k, v)| tracing::debug!("{} - {} total snippets after de-overlap", k, v.len()))
+        .flat_map(|(_, v)| v.into_iter().take(per_file_limit))
+        //.flat_map(|(_, v)| v.into_iter())
```
ggordonhall, line 185: Can we get rid of this?

ggordonhall (approved):

ggordonhall (approved):

PR #98, opened 2023-01-18
fix benchmarking on CI

nerdypepper:


llogiq:
Why not just `apt-get install protobuf-compiler` on the benchmark job, too?

nerdypepper:
i did try that, the protobuf version on `yum` repositories is too old. we could probably build it up ourselves and get the right version. however that is not the only missing dependency, we have to set up the boxes for our entire list of dependencies; including onnx etc. ideally i would like to do this with nix and keep all the versions in sync between local-production-benchmarks. 

self-hosted runners do something strange with the `PATH`, i can't seem to modify it during a run.

ggordonhall:
Closing this. To tackle as part of a wider review of our CI benchmarking

PR #101, opened 2023-01-18
Update documentation

ggordonhall:
Update README files with instructions on how to build and run bloop with natural language question answering. This is only a start; we'll be fleshing out the documentation in future updates.


apps/desktop/src-tauri/config.json
```diff
@@ -1,4 +1 @@
-{
-    "skip-indexing": true,
-    "github_client_id": "5620006101aa9914ab43"
-}
+{}
```
nerdypepper, line 1: can we make this config file optional?

server/README.md
```diff
@@ -24,25 +24,37 @@ Exuberant Ctags 5.8, Copyright (C) 1996-2009 Darren Hiebert
...
+
+```
+docker run -p 6333:6333 -p 6334:6334 \
+    -e QDRANT__SERVICE__GRPC_PORT="6334" \
+    qdrant/qdrant
+```
+
+You can run `bleep` without linking it to a `Qdrant` instance but calls to the `/answer` endpoint will return an error.
+
+You'll also need to run a local instance of `answer_api` (`bleep` expects it on port 7879 by default) which handles requests to the OpenAI API.
```
nerdypepper, line 38: can we link to the answer_api repo here?

nerdypepper (approved):
lgtm, just one comment.

server/README.md
ggordonhall: We need to open source first. Will do that and update after merging.

nerdypepper (approved):

apps/desktop/src-tauri/config.json
ggordonhall: It's not straightforward to optionally bundle additional files in Tauri (https://tauri.app/v1/guides/building/resources). Will stick this on the backburner.

PR #107, opened 2023-01-18
fix overlong chunks

llogiq:
The problem was two-fold. First, the last chunk ignored overlong lines (especially for files with one long line) and second, the  add_token_range method always extended the chunk back to the start of the line. So on an overlong line, the first partial chunk was OK, the next one a bit too long, the next one even longer and so on.

To make matters worse, onnxruntime broadcasted the dimension by 512, allocating huge swathes of memory in the process.

So here's hope that will at least significantly reduce our perf problem during indexing.

### DIFF
--- a/server/bleep/src/semantic/chunk.rs
+++ b/server/bleep/src/semantic/chunk.rs
@@ -177,7 +177,7 @@ fn add_token_range<'s>(
     if trimmed_end_byte <= start_byte {
         return;
     }
-    assert!(
+    debug_assert!(
         end - start < 256,
         "chunk too large: {} tokens in {:?} bytes {:?}",
         end - start,
--- a/server/bleep/src/semantic/chunk.rs
+++ b/server/bleep/src/semantic/chunk.rs
@@ -1,4 +1,7 @@
-use std::{fmt::{Display, Write}, ops::Range};
+use std::{
+    fmt::{Display, Write},
+    ops::Range,
+};
 
 use crate::text_range::{Point, TextRange};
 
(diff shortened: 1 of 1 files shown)


server/bleep/src/semantic/chunk.rs
```diff
@@ -172,6 +177,13 @@ fn add_token_range<'s>(
     if trimmed_end_byte <= start_byte {
         return;
     }
+    assert!(
```
ggordonhall, line 180: Rather than `assert` here, we could `warn`, and truncate the number of tokens produced by the tokenizer. In `semantic.rs` we can put something like:

```
 let mut tokenizer =
            tokenizers::Tokenizer::from_file(model_dir.join("tokenizer.json")).unwrap();
        tokenizer.with_padding(Some(PaddingParams {
            strategy: PaddingStrategy::BatchLongest,
            direction: PaddingDirection::Right,
            pad_id: 0,
            pad_type_id: 0,
            pad_token: "[PAD]".into(),
            pad_to_multiple_of: None,
        }));
```
when initialising the tokenizer.

ggordonhall (dismissed):
Looks good! Just added a thought

server/bleep/src/semantic/chunk.rs
llogiq: The problem with that is that, again, we don't have the actual number of tokens, and our calculation is merely a guess. In this case, the assert was there to ensure that we never under any circumstance get an overlong chunk. I'd be fine with using a `debug_assert!(..)`  instead.

ggordonhall: I suppose the problem with truncating at the tokeniser level is that we'd have to toggle truncation on-and-off depending on whether we are chunking or embedding. A better solution would be to not re-use tokens calculated when chunking, passing them to the embedder. We can address this in a future issue.

ggordonhall (approved):

PR #111, opened 2023-01-20
retry answer-api until success

nerdypepper:
continuously sends requests to answer-api until a successful response is received, or the max number of retries is hit. 

TODO: answer-api flattens openai responses to 500s in all cases, so the error message is supbar for the time being. with some improvements to answer-api, we can specialize the error message.

### DIFF
--- a/server/bleep/src/webserver/answer.rs
+++ b/server/bleep/src/webserver/answer.rs
@@ -2,7 +2,8 @@ use std::collections::HashMap;
 
 use axum::{extract::Query, response::IntoResponse, Extension, Json};
 use reqwest::StatusCode;
-use tracing::{error, info};
+use thiserror::Error;
+use tracing::{debug, error, info, warn};
 use utoipa::ToSchema;
 
 use crate::{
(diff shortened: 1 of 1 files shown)


server/bleep/src/webserver/answer.rs
```diff
@@ -350,6 +366,27 @@ impl<'s> AnswerAPIClient<'s> {
...
+            }
+            let response = self.send(prompt, max_tokens).await;
+            match response {
+                Ok(r) if r.status() == StatusCode::OK => break Ok(r),
+                Err(e) => break Err(AnswerAPIError::Fatal(e)),
+                _ => (),
+            };
+            warn!(%attempt, "answer-api returned {} ... retrying", response.unwrap().status());
+            attempt += 1;
+        }
```
llogiq, line 388: Why not use a ` for attempt in 0..self.max_attempts { .. }`  loop here?

llogiq (approved):
Just a small nit; I'll leave the code style to your discretion.

server/bleep/src/webserver/answer.rs
nerdypepper: sure, that would clean the extra `if` check.

PR #114, opened 2023-01-20
Replace segment with rudderstack on the backend

nerdypepper:
- renames the `segment` module to `analytics`
- slightly modifies the data schema, to represent the different stages of the query pipeline
- requires two new environment variables: `ANALYTICS_KEY_BE` and `ANALYTICS_DATA_PLANE`, these values can be found in the rudderstack dashboard and must be set in the `.env` file


server/bleep/src/lib.rs
```diff
@@ -281,14 +282,17 @@ impl Application {
                 Err(e) => bail!(e),
             };
 
-        let segment = if let Ok(segment_key) = std::env::var("SEGMENT_KEY_BE") {
-            info!("initializing segment");
-            Some(Segment::new(segment_key))
+        let analytics_client = if let (Ok(key), Ok(data_plane)) = (
+            std::env::var("ANALYTICS_KEY_BE"),
```
ipauler, line 286: should we update github workflow ?

nerdypepper: these are injected at runtime rather than build time, similar to how `segment` operated. i wonder if this is better off injected during build time? cc @ggordonhall @rsdy.

nerdypepper: the tauri-release flow needs to be updated. https://github.com/BloopAI/roadmap/issues/377

PR #89, opened 2023-01-15
Allow permission management through authenticating users

rsdy:
This PR adds a mode for `bloop` to work as a [Github App](https://docs.github.com/en/developers/apps/getting-started-with-apps/about-apps) installation that's limited in scope to [a single organization](https://docs.github.com/en/developers/apps/building-github-apps/authenticating-with-github-apps#authenticating-as-an-installation).

The base assumption is that the **only** entity that controls the Github App keys is the auth bastion server (`auth_server_url` command line/config option), while **only** the bloop instances hold the users' OAuth keys.

This PR holds an initial bloop client implementation that talks to an emergent auth bastion specification (TBD).

## Current status:

 * the main addition is `webserver/aaa.rs` to hold AAA-related functionality
 * there are lot of corner cases not covered here, I tried to document these in the code as I went
 * wiring up bloop to an auth bastion should work, and syncing repos should be possible (not tested, though)
 * users should be able to sign up with the correct Github App setup, and an auth bastion
 * token issuance
 * key rotation, check for expiry, upstream revokation
 
## Not implemented

 * setting up a root admin user (should send a `/private/hello` to the auth bastion to see who asked for the deployment?)
 * persistence of user oauth keys (this forces users to re-login every time we restart the service)
 * there are a few DoS opportunities
 * error handling, nice HTTP responses
 * `TODO` comments in the PR code, `todo!()` macros
 * the auth bastion server itself

PR #115, opened 2023-01-21
Add `PrivateServer` mode, implement GitHub App API integration

calyptobai:
This PR supersedes #89, adding working support for GitHub App installations. Some changes are included:

- The need for a new auth server is no longer necessary, so we do everything in `bleep`. GitHub's API allows us to set the Callback URL to `https://bloop.ai`, and we can set the `redirect_uri` parameter in the OAuth API to a subdomain
- As discussed offline, the GitHub app will first be installed into the organization before the bloop instance is provisioned, avoiding the need for any webhooks (a post-install URL is used instead).
- Fine-grained authentication is removed. It can be rolled back by using the original commits in #89. This is better left to another PR as there are many questions brought up offline, and this should not block merging initial GitHub App functionality

TODO:
- [x] This PR is missing a final bit of functionality; checking whether the user belongs to the installation organization. This can be done with the use of a single API call (`/orgs/{org_name}/members/{user_name}`)
- [x] Fix remaining TODO items

---

Co-authored-by: rsdy <p@symmetree.dev>


server/bleep/src/webserver/aaa.rs
```diff
@@ -0,0 +1,263 @@
...
+        "https://github.com/login/oauth/authorize\
+         ?client_id={client_id}\
+         &state={state}\
+         &allow_signup=false\
+         &redirect_uri={redirect_uri}",
+    );
+
+    auth_layer.initialized_login.insert(state);
+
+    Redirect::to(github_oauth_url)
```
anastasiya1155, line 79: I don't think redirects will work for us, we should find another way of doing it. When a client makes a request to this endpoint and gets a redirect in response it tried to make an API call to this address, but it should open it as a webpage in a new tab instead. Also it won't work in Tauri app.

server/bleep/src/webserver/aaa.rs
```diff
@@ -0,0 +1,263 @@
...
+    Extension(auth_layer): Extension<Arc<AuthLayer>>,
+    Query(params): Query<AuthorizedParams>,
+    jar: CookieJar,
+) -> impl IntoResponse {
+    let AuthorizedParams { state, code } = params;
+
+    auth_layer
+        .initialized_login
+        .remove(&state)
+        .expect("invalid state key");
```
anastasiya1155, line 126: I don't understand why but I always get this error although I checked that the state param is the same

rsdy:
There are multiple issues with this PR.

1. My original work is misattributed in the commits.
2. The considerations for this work *seem to be* that the original brief’s deployment model was simplified, but this is only my conclusion after review.
3. It essentially removes all sense of security from the original PR.

This work removes all kinds of authentication of user-provided credentials, it is trivial for users to impersonate any organization’s any user in possession of their GitHub OAuth credentials issued by *any* application. This is made especially worse, since token expiration is unauthenticated, therefore a *fired* employee could at any point log back in to bloop, and extract data by simply spoofing the expiry time.

It is also concerning that the PR’s token system unapologetically uses GitHub tokens to authenticate for Bloop, and stores these unencrypted on users’ boxes. One of the primary threats to developer work is [credential theft](https://metacurity.substack.com/p/bitcoin-core-developer-lost-basically), and ANY packaging system, including Python, Node.js, or Cargo could host packages that steal these credentials build-time.

While the tokens currently are limited in scope, they do allow for impersonation and the design exposes any subsequent work to potentially leaking tokens with extended scopes.

The solution to authorization seems to be based on the expiration time of the GitHub-issued tokens. This is neglectful, and doesn’t actually log users out when they revoke access through GitHub. Since we don’t control GitHub’s HMAC keys, the only way to actually validate the tokens would be through in-band checks on every request, which is a huge performance impact. Since GitHub's OAuth doesn't use public key cryptography, this issue cannot be fixed without the introduction of an extra layer of security.

These issues are architectural, and make this work unfit for purpose. I think it would be worthwhile adapting the original PR to the new deployment model instead of continueing this work, as #89 provides a complete  authentication/authorization solution that doesn't suffer from either of these flaws.

calyptobai:
Reopening as per discussion with @rsdy and @stunningpixels. Summary of remaining work:

- Complete missing TODO items in PR description and in code
- Encrypt cookies

server/bleep/src/webserver/aaa.rs
calyptobai: Perhaps you called this route multiple times? It should only work once, during a redirect from the GitHub login page. Future calls will fail as the key is removed.

server/bleep/src/webserver/aaa.rs
calyptobai: Updated to return `{ "oauth_url": "https://..." }` to make it easier.

anastasiya1155: thanks

server/bleep/src/webserver/aaa.rs
anastasiya1155: could be

server/bleep/src/webserver/aaa.rs
```diff
@@ -0,0 +1,341 @@
...
+        .send()
+        .await
+        .unwrap()
+        .json()
+        .await
+        .unwrap();
+
+    (
+        jar.add(AuthCookie::new(gh_token).to_cookie()),
+        Redirect::to("/"),
```
anastasiya1155, line 185: can you also remove this redirect and instead create an endpoint that will just return true if the user is logged in, so that on the FE we know if it's time to show the next step?

server/bleep/src/lib.rs
```diff
@@ -380,40 +446,117 @@ impl Application {
...
+        }
+
+        let auth =
+            remotes::github::Auth::from_installation(installation, install_id, octocrab).await?;
+        let credential = BackendCredential::Github(auth);
+        self.credentials.insert(Backend::Github, credential.clone());
+
+        Ok(credential)
+    }
+
```
rsdy, line 542: Remove these from `Application`.

server/bleep/src/webserver/aaa.rs
```diff
@@ -0,0 +1,341 @@
...
+    ///
+    /// Maps to the time this login attempt was created.
+    initialized_login: DashMap<State, Instant>,
+
+    /// The HTTP client.
+    client: reqwest::Client,
+}
+
+impl AuthLayer {
+    fn clean_old_states(&self) {
```
rsdy, line 219: This is not sufficient against DoS protection, as your expiration is time-based, but your memory can be filled up within 5 minutes with login requests.

rsdy (changes requested):
If you entirely rely on cookies for auth, you'll probably want to implement CSRF protection as well.

server/bleep/src/webserver/aaa.rs
calyptobai: Considering we have to apply rate limiting at the deployment level, this isn't an issue specific to this code. With rate limiting, it's impossible for a single actor to hammer the `state` cache. Multiple actors involves a DDoS, which needs to be mitigated elsewhere as this can flood us at the network level.

server/bleep/src/webserver/aaa.rs
calyptobai: This is the redirect target for GitHub's login page; after logging in we hit this route, which ultimately redirects to `/`. It's not meant to be called directly as part of an API call.

server/bleep/src/webserver/aaa.rs
rsdy: Unfortunately this assumption doesn't hold in an environment we do not control, such as on premise. A sufficient solution to this DoS would be to limit the number of allowed parallel outstanding login flows to something reasonably large, eg 1024, and cancel anything more than that.

You're correct, DDoS protection can't be in scope here.

calyptobai: This is not worth spending time on here. We have the same issues with basic HTTP attacks being able to jam the server because we're using axum without a more robust server in front.

PR #116, opened 2023-01-23
use gpt2 for token counting

nerdypepper:
this change was somehow reverted with the `semantic` merge.

### DIFF
--- a/server/bleep/src/semantic.rs
+++ b/server/bleep/src/semantic.rs
@@ -223,7 +223,7 @@ impl Semantic {
             repo_name,
             relative_path,
             buffer,
-            &self.tokenizer,
+            &self.gpt2_tokenizer,
             self.config.max_chunk_tokens,
             15,
             self.config


llogiq (commented):
Do we have the GPT2 tokenizer in use for embedding already?

nerdypepper:
@llogiq no we don't. this change only makes the token counting calculations accurate (using gpt2 for both chunking as well as answer-api). i can't really speak for the improvement (if any) in using gpt2 for embedding, that might be worth discussing.

llogiq:
But if we have two different tokenizers, the chunking calculation may be off.

nerdypepper:
ah yes, every chunk needs to limit itself to 256 MiniLM tokens, but at the same time, we need to know how many GPT2 tokens are in a chunk.

llogiq:
I guess it won't make that much of a difference, probably truncating before embedding might be a good idea though.

ggordonhall (commented):
@nerdypepper When chunking we want to make sure we get as close as possible to 256 _MiniLM_ tokens so we need to use `self.tokenizer` at this stage. Why would we need to calculate the number of GPT tokens here? Isn't that something we can do on the fly after retrieval?

nerdypepper:
@ggordonhall so some of the calculations at answer-api depend on having every chunk be the same size; of 256 GPT2 tokens. namely, the number of snippets is locked to 13; since 256 * 13 is ~3300, leaving us enough space for the rest of the prompt + completion. we can rework that to be more dynamic; but we would have to tokenize once more with GPT2 to know how big every chunk is.

ggordonhall:
@nerdypepper How much of an overhead does query-time GPT2 tokenization add? My hunch is that it's negligible compared to the time it takes to generate an answer, especially if we use Tiktoken rather than Huggingface.

nerdypepper:
it is largely negligible.

PR #117, opened 2023-01-24
experiment: unify Rust caches

llogiq:
This just adds a shared key to all our workflows. Not completely sure if that works correctly, but we cannot check before pushing this.


llogiq:
This will obviously build as slow as other builds, because it cannot reuse prior caches.

rsdy (approved):
Let's try...

PR #118, opened 2023-01-24
Update prompt to discourage model hallucination

ggordonhall:
Empirically, this prompt reduces the chance that the model hallucinates code and passes it off as belonging to the indexed repo. Also removes the instruction to return the file that the answer was derived from. This is always the path of the first entry in `snippets`, so we just add this to the response object and remove the redundant `index` field (which was always 0).


nerdypepper:
i would suggest that the indents be removed, the tend to encourage indents in the response as well.

PR #120, opened 2023-01-24
change chunking algorithm to reduce small chunks

llogiq:
The reason we had a suboptimal chunking strategy was that we always took the next-newline before max_tokens, whether that was early or not, and didn't stop when reaching `the end.

Now I have changed the algorithm a good deal to chunk on newlines only when it's within 3/4s of the max_token size, and try to use word boundaries if that doesn't work.


rsdy (approved):

ggordonhall:
> 

Tracking test coverage in another ticket

PR #119, opened 2023-01-24
Add Docker caching to backend and frontend builds

rsdy:
This is supposed to give us a caching layer for our dependencies, which should be a big win.

An extra step would be to store our built Docker artifacts & layers in the [GitHub Container Registry](https://github.com/docker/build-push-action/), so we could rely on the CI always keeping a copy of `main` up-to-date.

PR #121, opened 2023-01-24
Serve front-end in `PrivateServer` mode

calyptobai:
This adds a new option `--frontend-dist` which accepts a path to serve as part of the `PrivateServer` mode. All API routes have been moved to `/api`, and the front-end `baseURL` has been adjusted accordingly.

### DIFF
--- a/server/bleep/src/lib.rs
+++ b/server/bleep/src/lib.rs
@@ -140,6 +140,14 @@ impl Environment {
             PrivateServer => false,
         }
     }
+
+    pub(crate) fn use_aaa(&self) -> bool {
+        matches!(self, Self::PrivateServer)
+    }
+
+    pub(crate) fn serve_frontend(&self) -> bool {
+        matches!(self, Self::PrivateServer)
+    }
 }
 
 #[derive(Deserialize, Parser, Debug)]
--- a/server/bleep/src/lib.rs
+++ b/server/bleep/src/lib.rs
@@ -140,16 +140,6 @@ impl Environment {
             PrivateServer => false,
         }
     }
-
-    /// Use AAA layer (authentication, authorization, accounting)
-    pub(crate) fn use_aaa(&self) -> bool {
-        use Environment::*;
-        match self {
-            Server => false,
-            InsecureLocal => false,
-            PrivateServer => true,
-        }
-    }
 }
 
 #[derive(Deserialize, Parser, Debug)]
@@ -250,6 +240,10 @@ pub struct Configuration {
     #[clap(long)]
     /// Chunking strategy
     pub overlap: Option<OverlapStrategy>,
+
+    /// Path to built front-end folder
+    #[clap(long)]
+    pub frontend_dist: Option<PathBuf>,
 
     //
     // External dependencies
--- a/server/bleep/src/lib.rs
+++ b/server/bleep/src/lib.rs
@@ -119,7 +119,26 @@ fn default_max_chunk_tokens() -> usize {
 pub enum Environment {
     /// Safe API that's suitable for public use
     Server,
-    /// Access to the API is access controlled using an OAuth provider
+    /// Use a GitHub App installation to manage repositories and user access.
+    ///
+    /// Running the server in this environment makes use of a GitHub App in order to list and fetch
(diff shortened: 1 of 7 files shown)


server/bleep/src/webserver.rs
```diff
@@ -65,33 +66,45 @@ pub async fn start(app: Application) -> Result<()> {
...
-        .with_state(app)
+        .with_state(app.clone())
         .layer(CorsLayer::permissive())
         .layer(CatchPanicLayer::new());
 
+    let mut router = Router::new().nest("/api", api);
+
+    if let Environment::PrivateServer = app.env {
+        router = router.merge(SpaRouter::new(
+            "/",
```
anastasiya1155, line 100: we need more than the root route here, if the app is showing /search page it results in 404. We need all endpoints except for /api to serve FE

server/bleep/src/remotes/poll.rs
```diff
@@ -26,7 +26,7 @@ const POLL_INTERVAL_MINUTE: &[Duration] = &[
 
 pub(crate) async fn check_credentials(app: Application) {
     loop {
-        if app.env.use_aaa() {
+        if let Environment::PrivateServer = app.env {
```
rsdy, line 29: Don't do direct matching of these. Check for individual permissions bound to the server mode, as it was originally, this helps us understand the different modes the app can run in easily.

server/bleep/src/webserver.rs
```diff
@@ -65,33 +66,45 @@ pub async fn start(app: Application) -> Result<()> {
...
 
     if app.env.allow_path_scan() {
-        router = router.route("/repos/scan", get(repos::scan_local));
+        api = api.route("/repos/scan", get(repos::scan_local));
     }
 
     // Note: all routes above this point must be authenticated.
-    if app.env.use_aaa() {
-        router = aaa::router(router, app.clone());
+    if let Environment::PrivateServer = app.env {
```
rsdy, line 79: Same here. Revert this change.

server/bleep/src/webserver.rs
```diff
@@ -65,33 +66,45 @@ pub async fn start(app: Application) -> Result<()> {
...
         .layer(Extension(app.semantic.clone()))
         .layer(Extension(app.clone()))
-        .with_state(app)
+        .with_state(app.clone())
         .layer(CorsLayer::permissive())
         .layer(CatchPanicLayer::new());
 
+    let mut router = Router::new().nest("/api", api);
+
+    if let Environment::PrivateServer = app.env {
```
rsdy, line 98: And here you should introduce a `app.env.serve_frontend()` boolean.

rsdy (changes requested):

server/bleep/src/webserver.rs
```diff
@@ -65,33 +66,45 @@ pub async fn start(app: Application) -> Result<()> {
...
+        router = router.merge(SpaRouter::new(
+            "/",
+            app.config
+                .frontend_dist
+                .as_ref()
+                .expect("need path to built frontend folder"),
```
rsdy, line 104: I don't think this should be a hard error, just continue if not found. We could have an API running separately from the frontend assets due to eg. CDN deployments. A hard error makes this implementation dependent instead of configuration dependent.

rsdy:
See #119 for packaging

server/bleep/src/lib.rs
```diff
@@ -141,14 +141,12 @@ impl Environment {
...
-        match self {
-            Server => false,
-            InsecureLocal => false,
-            PrivateServer => true,
-        }
+        matches!(self, Self::PrivateServer)
+    }
+
+    pub(crate) fn serve_frontend(&self) -> bool {
+        matches!(self, Self::PrivateServer)
```
llogiq, line 149: Do we need two methods for the same thing here?

llogiq (approved):
Just a very small nit code-wise. Otherwise, some doc comments (what a private server is (for), how to set it up) would be helpful.

server/bleep/src/lib.rs
calyptobai: @rsdy requested this, I'll revert the change

PR #125, opened 2023-01-25
Remove FE env prefix

ipauler:


anastasiya1155 (approved):
can you also update .env.example?

apps/desktop/.env.example
```diff
@@ -1,6 +1,6 @@
-VITE_ANALYTICS_WRITE_KEY_PROD=
```
ggordonhall, line 1: Are these env vars just used by the Tauri app? if not then we should move this up to the top level of the repo.

ggordonhall (approved):

apps/desktop/.env.example
ipauler: moved all env to root

PR #126, opened 2023-01-25
Self serve FE

anastasiya1155:


apps/desktop/src-tauri/src/backend.rs
```diff
@@ -26,7 +26,7 @@ where
 
     let app = app.handle();
     tokio::spawn(async move {
-        let initialized = Application::initialize(Environment::InsecureLocal, configuration).await;
+        let initialized = Application::initialize(Environment::PrivateServer, configuration).await;
```
anastasiya1155, line 29: this needs to be conditional, but not sure how to do it

client/src/main.tsx
```diff
@@ -21,6 +21,9 @@ ReactDOM.createRoot(document.getElementById('root') as HTMLElement).render(
         },
         invokeTauriCommand: () => {},
         release: packageJson.version,
+        apiUrl: import.meta.env.VITE_API_URL || '',
+        isRepoManagementAllowed: false,
```
anastasiya1155, line 25: this should depend on user role/permissions

ipauler (approved):

PR #131, opened 2023-01-26
Build Docker images on merging to `main`

rsdy:
Caching to GHA caches works, but due to the sheer size, potentially conflicts with with rust-cache from other workflows.

Still investigating this


ggordonhall (commented):
We could set them to build on manual trigger? We can wait it out when we need them, but don't need to run every PR into main.

llogiq (dismissed):
Makes sense for now.

PR #133, opened 2023-01-26
add chunk docs & tests

llogiq:
Also speed up the whole thing by giving the minimum chunk size directly to `by_tokens`  instead of filtering after the fact, and set the number of onnxruntime threads to ` NUM_OMP_THREADS`  even when building without OpenMP.

This closes https://github.com/BloopAI/roadmap/issues/383.

### DIFF
--- a/server/bleep/src/semantic.rs
+++ b/server/bleep/src/semantic.rs
@@ -104,6 +104,12 @@ impl Semantic {
                 .build()?,
         );
 
+        let threads = if let Ok(v) = std::env::var("NUM_OMP_THREADS") {
+            str::parse(&v).unwrap_or(1)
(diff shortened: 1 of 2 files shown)


ggordonhall (approved):
Docs look good. Are we using ONNX Runtime binaries that were built by Microsoft? Can see here https://github.com/pykeio/ort/blob/bb2f9240e7018ab86bc7dba9925f1a59ad4b1f62/src/session.rs#L157 that the runtime will read `OMP_NUM_THREADS` directly and ignore ` with_intra_threads`

llogiq:
Currently we build the runtime statically, which is why I added that little snippet of code to keep the function alive.

PR #135, opened 2023-01-26
Move all configuration items to be provisioned from the config

rsdy:
Moving everything to a single interface simplifies Dockerized deployments, as we only need to generate a single configuration with all values included.

For development purposes, this practically means we might want to gradually move away from command line arguments altogether in favour of loading the configuration file, similar to how `qdrant` also works.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -384,7 +384,6 @@ dependencies = [
  "criterion",
  "dashmap",
  "directories",
- "dotenvy",
  "dunce",
  "either",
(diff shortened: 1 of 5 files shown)


ggordonhall (approved):
Really like moving the `Configuration` struct out of `lib.rs`, I think the rest lgtm

nerdypepper (approved):
this PR cancels this task: https://github.com/BloopAI/roadmap/issues/377

PR #136, opened 2023-01-27
split workflow caches

llogiq:
In my recent investigations, I found that giving all workflows the same cache will lead to errors where one workflow takes away the cache from another. So let's try splitting the caches instead.

Related to https://github.com/BloopAI/roadmap/issues/371

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -31,7 +31,8 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
         with:
-          shared-key: "bloop"
+          prefix-key: "bloop"
+          shared-key: "doc"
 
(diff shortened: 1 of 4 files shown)


llogiq:
Again, this is going to build slowly, because we set up new cache keys, and won't be able to re-use the old caches.

rsdy (approved):
What's a good way to test if caching actually works? Re-running the jobs would pick up the caches I think?

llogiq:
> What's a good way to test if caching actually works? Re-running the jobs would pick up the caches I think?

Unfortunately, that's not enough. We need to look across multiple PRs to see if the caches get picked up, and then we should look into whether the caches as set up actually help or if we need an additional component like sccache.

PR #141, opened 2023-01-27
Unify cache dirs for Tauri app

rsdy:
Also restrict internal functions to prevent future accidental misuse, such as what lead to this.


ggordonhall (approved):
Nice

PR #142, opened 2023-01-27
Limit maximum number of parallel authentication sessions

rsdy:
This caps the max memory used by pending authentications, mitigating the DoS here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -422,6 +422,7 @@ dependencies = [
  "tantivy",
  "tempdir",
  "thiserror",
+ "time 0.3.17",
  "tokenizers",
  "tokio",
  "tower",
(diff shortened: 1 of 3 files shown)


calyptobai (approved):
LGTM :+1: though worth noting this doesn't quite mitigate an authentication DoS attack here.

For example:

```
while true; do curl "$SERVER_URL/auth/login/start"; done
```

This would previously would fill up server memory, but now instead continuously fills up the maximum 512 `initialized_login` limit and blocks other users from signing on.

There are more DoS vulnerabilities in the webserver like this, e.g. the above bash loop making a large number of requests to `/q`. Rather than solving this at the app level, which would still leave us vulnerable to e.g. network-layer flooding, this would be better solved at the deployment level by setting up a rate limit, to make use of a cloud provider's high bandwidth limiting systems.

PR #144, opened 2023-01-28
Stream OpenAI results

calyptobai:
This PR adds support for streaming OpenAI results for query explanations. This changes the API for `/answer` - the response is now streamed using HTTP SSE. For the first message, a JSON object is sent with the same data as before, but without the explanation text. Subsequent messages are incoming tokens, as streamed from OpenAI.

Response schema changes: (CC @anastasiya1155)

```diff
 struct AnswerResponse {
     user_id: String,
     query_id: uuid::Uuid,
     snippets: Vec<Snippet>,
-    selection: api::Response {
-       answer_path: String,
-       explanation: String,
-    }
+    answer_path: String,
  }
```

Note that we no longer return `explanation` as part of the `AnswerResponse`. The explanation field is streamed directly in the form of plaintext SSE messages which can be concatenated together to form the final explanation string.


server/bleep/src/webserver.rs
```diff
@@ -128,19 +127,17 @@ where
     Json(Response::from(val))
 }
 
-pub(in crate::webserver) fn error<'a>(
-    kind: ErrorKind,
-    message: impl Into<Cow<'a, str>>,
-) -> Json<Response<'a>> {
+pub(in crate::webserver) type Result<T, E = Error> = std::result::Result<T, E>;
```
llogiq, line 130: Did you know that `pub(in ..)` might be deprecated in the 2024 edition?

llogiq (approved):

server/bleep/src/webserver/answer.rs
```diff
@@ -249,54 +244,58 @@ pub async fn handle(
...
     let answer_path = snippets.get(0).unwrap().relative_path.to_string();
 
-    Ok::<_, Json<super::Response<'static>>>(Json(super::Response::Answer(AnswerResponse {
-        snippets,
-        query_id,
-        user_id: params.user_id,
-        selection: api::Response {
-            answer: snippet_explanation,
+    let initial_event = Event::default()
+        .json_data(super::Response::Answer(AnswerResponse {
```
nerdypepper, line 256: can we implement `Into<super::Response> for AnswerResponse`, like we do for all these response types? like how `HoverableResponse` and `TokenInfoResponse` convert to `super::Response`. we could every macro-ify this impl in `webserver.rs`.

nerdypepper (approved):

server/bleep/src/webserver.rs
calyptobai: I didn't know that 😄 It's not particularly useful here, I'm just following the visibility specifier that other types use.

server/bleep/src/webserver/answer.rs
```diff
@@ -452,11 +513,20 @@ Answer in GitHub Markdown:",
         prompt
     }
 
-    async fn select_snippet(&self, prompt: &str) -> Result<reqwest::Response, AnswerAPIError> {
-        self.send_until_success(prompt, 1, 0.0).await
+    async fn select_snippet(&self, prompt: &str) -> super::Result<String> {
```
ggordonhall, line 516: We could also maintain the non-streaming `answer-api` endpoint for these single token responses.

ggordonhall (approved):
Looks good, @anastasiya1155 are you adding the frontend changes to this branch?

llogiq (approved):

PR #147, opened 2023-01-30
Allow unlinking Github user profile from local app

rsdy:
I think this is all we can do here.

According to Github docs, when [deauthorizing a token](https://docs.github.com/en/rest/apps/oauth-applications?apiVersion=2022-11-28#delete-an-app-token), they require both a `client_id`, AND a `client_secret`, which we decidedly don't have since we're using the device flow.

Octocrab also doesn't seem to have corresponding functionality. So we nuke the key and move on.


anastasiya1155 (approved):
Will this work for self-serve too?

rsdy:
No, there's no revocation for GH tokens on self serve at the moment. Likely we'll want another endpoint to do that.

ggordonhall (approved):
Lgtm

PR #150, opened 2023-01-31
change cross-file heuristic to be smarter

nerdypepper:
previously on scope-graphs; the heuristic used to fetch repo-wide ref/def data was as follows:

> if the cursor is currently hovering over a ref, fetch all defs in this repo with the same name, and filter out ones that are probably not the right def (i.e., local variable defs are mostly not the right def)

the issue with this heuristic is that it leaks the idea of a "symbol kind" to something that should be agnostic of such information. 

the new method proposed, ignores the idea of interleaving "symbol kinds" and scopes. it works as follows:

> top-level definitions are the only definitions reachable across file boundaries

this works well because it creates a new latent scoping rule: every file creates a scope. consider the following rust codebase:

```
src/
    `- a.rs
    `- b.rs
```

where:

```rust
// a.rs

fn a_main() {
    let a_foo = 1;
}
```

```rust
// b.rs
use a::a_main;
fn b_main() {
    a_main();
    a_foo + 1;
}
```

the scope-graphs generated, might look like so (defs in green, refs in yellow):

graphs_rot

the scope-resolution for the reference `a_main` (in `b.rs`) exists, and looks like so:
resolution_rot

and the scope-resolution for the reference `a_foo` (in `b.rs`) fails, because there is no top-level definition of `a_foo` anywhere in the codebase:
no_resolution_rot


nerdypepper:
this PR also includes a small fix to rust: module declarations were not creating scopes.

ggordonhall (approved):
This makes sense, and should cut out may of the false positives that we had before doing cross-file scope-graph resolution. This should work particularly well for a language like Rust, but I imagine we'll lose most of the cross-file support for a language like Java. We'll probably have to wait until we have stack graph support to really handle well.

PR #151, opened 2023-01-31
Silence invalid lang warnings

ggordonhall:
After https://github.com/BloopAI/bloop/pull/88 we pass directory paths to the `File` index worker. Warnings are thrown when we check these dir paths for language. This PR only looks up language for file paths.


server/bleep/src/indexes/file.rs
```diff
@@ -498,15 +498,18 @@ impl File {
...
-        let lang_str = repo_info
-            .langs
-            .path_map
-            .get(&entry_disk_path)
-            .unwrap_or_else(|| {
-                warn!("Path not found in language map");
-                &Some("")
-            })
-            .unwrap_or("");
+        let lang_str = match &entry_disk_path.is_file() {
```
rsdy, line 501: It's really a nit, but I'd phrase this as an `if` expression.

rsdy (approved):

PR #153, opened 2023-01-31
Handle disconnect/remote repo deletion

rsdy:
If syncing a remote repository fails due to credential issues, or it seems like the remote repository is no longer available, we mark it `RemoteRemoved`, a zombie state. This is to notify the user that there has been a potentially fatal error here.

Until further user action, there's no re-indexing of the repository whatsoever, but we don't immediately clean everything up. A follow-up API action will be necessary for a proper cleanup of these zombie artifacts and indexes.

Until the user actively chooses to remove the remote repository, it remains searchable on their bloop instance.

Note that according to the [Github API docs](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28) we never actually receive a "Permission Denied" as long as the rate limiting is not hit, and credentials are valid, ie. we only get 404 if repository or organization access is revoked. 

@anastasiya1155 It would be helpful if you could integrate the required frontend signaling into this PR.

Note: we already have a `DELETE /api/repos/indexed/<reporef>` endpoint to actually delete these repositories.


llogiq (approved):
I think the code might use some tests, but that was the case before this PR already.

Otherwise looks good.

ggordonhall (commented):
Tested and I get a `GenericError` status for repos that have been removed:
```
{
  "list": [
    {
      "provider": "github",
      "name": "ggordonhall/bench",
      "ref": "github.com/ggordonhall/bench",
      "local_duplicates": [],
      "sync_status": {
        "error": {
          "message": "low-level code: GenericError"
        }
      },
      "last_update": "2022-10-19T13:22:22Z",
      "last_index": "2023-02-06T22:04:02Z",
      "most_common_lang": "TOML"
    }
  ]
}
```

server/bleep/src/background.rs
```diff
@@ -213,17 +223,36 @@ impl IndexWriter {
...
         &self,
         repo: &Repository,
         writers: &indexes::GlobalWriteHandleRef<'_>,
     ) -> anyhow::Result<()> {
         let IndexWriter(Application { config, .. }) = self;
 
         repo.delete_file_cache(&config.index_dir)?;
+        if !matches!(repo.remote, RepoRemote::None) {
+            tokio::fs::remove_dir_all(&repo.disk_path).await?;
+        }
```
llogiq, line 255: Perhaps add a comment that this does the actual cleanup of deleted repos. Also are we sure we want to delete the local workspace? Should we put this behind an option?

ggordonhall (approved):

server/bleep/src/background.rs
rsdy: This doesn't delete any local repos, only repos managed by bloop itself. It also only cleans up the indexes/state when the user explicitly flips it into `Removed`, otherwise we'll keep these around.

Not sure what you mean by putting it behind an option?

PR #155, opened 2023-02-01
Lock cargo CI workflows

rsdy:
This adds `--locked` to all automated CI workflows in hope of throwing away caches less often.


llogiq (approved):
That rate limit sure gets on my nerves.

Otherwise this is also a good step towards reproducible builds. :+1:

PR #160, opened 2023-02-02
Don't start bloop if `qdrant_url` is invalid

rsdy:
This change makes `qdrant_url` a **required argument** if you want semantic search to function when running bloop as a backend!

The default configuration in the desktop app continues to use semantic, as expected.


server/bleep/src/lib.rs
```diff
@@ -103,15 +102,17 @@ impl Application {
...
+                match Semantic::initialize(&config.model_dir, url, Arc::clone(&config)).await {
+                    Ok(semantic) => Some(semantic),
+                    Err(e) => {
+                        bail!("Qdrant initialization failed: {}", e);
+                    }
                 }
-                Err(e) => bail!(e),
-            };
+            }
+            None => None,
```
ggordonhall, line 114: We should log here if Qdrant isn't initialised, but bloop startup continues.

ggordonhall (approved):

server/bleep/src/lib.rs
llogiq: Yeah, simply working without semantic search without so much as a warning would be bad UX.

llogiq (approved):
Fine apart from losing the warning on start without qdrant.

Perhaps also write a comment to outline the behavior (accept launching without qdrant URL, check if given).

llogiq (approved):

PR #162, opened 2023-02-03
Use self-hosted runner for docker build

rsdy:
Build without caches: 25m
Build with caches: 16m (lots of cache misses, likely we can do better)
Build without changes: 2m


.github/workflows/build-on-pr.yml
```diff
@@ -1,4 +1,4 @@
-name: Build&Push API docker container image with SHA tag
+name: Build docker container
```
ggordonhall, line 1:
```suggestion
name: Build and push docker container
```

ggordonhall (approved):
Would be good to also run the test workflow on a self-hosted runner. We've had trouble with caching there too.

PR #167, opened 2023-02-06
Add bot authentication flow

calyptobai:
This is a very simple implementation that leaves token generation to the deployment step.


rsdy (approved):
What's the slack-side of the authentication flow here? This is probably fine for just Slack, but I don't think this will scale very well to other bots.

ipauler:
@calyptobai we can use `bot_secret` insteand of `slackbot_secret` since this will be generic for multiple bot integrations

calyptobai:
> What's the slack-side of the authentication flow here? This is probably fine for just Slack, but I don't think this will scale very well to other bots.

This is only a token intended to be easy to generate in the deployment phase, specifically so that our Slack bot can connect to he server. It's not a general solution, it remains to be seen what that might look like.

PR #168, opened 2023-02-06
Error refactor

calyptobai:
This allows us to use simple `Result<impl IntoResponse>` for webserver handler return types, and use plain `?`-based error handling, which correctly propagates HTTP status codes.

### DIFF
--- a/server/bleep/src/webserver/answer.rs
+++ b/server/bleep/src/webserver/answer.rs
@@ -6,9 +6,8 @@ use std::{
 
 use axum::{
     extract::Query,
-    http::StatusCode,
     response::{sse::Event, IntoResponse, Sse},
-    Extension, Json,
+    Extension,
 };
 use futures::{Stream, StreamExt, TryStreamExt};
 use thiserror::Error;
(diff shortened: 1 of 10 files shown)


server/bleep/src/webserver/answer.rs
```diff
@@ -95,36 +94,21 @@ const SNIPPET_COUNT: usize = 13;
...
-            super::error(ErrorKind::User, e.to_string()),
-        )
-    })?;
-    let target = query.target().ok_or_else(|| {
-        (
-            StatusCode::BAD_REQUEST,
-            super::error(ErrorKind::User, "missing search target".to_owned()),
-        )
-    })?;
+) -> super::Result<impl IntoResponse> {
```
rsdy, line 97: This may look better included in `prelude` instead of relying on typing out `super::Result` at every use?

rsdy (approved):
This looks pretty good. Major improvements everywhere.

server/bleep/src/webserver/file.rs
```diff
@@ -2,12 +2,11 @@ use std::sync::Arc;
...
 use axum::{
     extract::{Path, Query},
-    response::{IntoResponse, Result},
+    response::IntoResponse,
     Extension,
 };
-use reqwest::StatusCode;
 
-use super::{json, prelude::Indexes, EndpointError, ErrorKind};
+use super::{json, prelude::Indexes, Error};
```
nerdypepper, line 9: would you mind cleaning up all the imports to make use of `prelude`? this could be `use crate::webserver::prelude::*`.

PR #171, opened 2023-02-07
index less

llogiq:
If my hunch is right, the recent indexing slowdown is due to the fact that we also index a full onnxruntime (because we don't filter out the `target`  directory. This should hopefully help.


rsdy (approved):

ggordonhall:
> if the directory is present in `.gitignore` it shouldn't affect index times. does the `target` directory somehow have a git repository inside it?

It does, and it was being indexed, but that has already been addressed here: https://github.com/BloopAI/bloop/pull/148. I don't think we need to add any special logic to handle `target/`. I do think that excluding `.lock` files is a good should though.

ggordonhall (approved):

PR #174, opened 2023-02-08
references files from across the repo in the prompt

nerdypepper:
cross-referencing is a three-staged process:

- pass the original snippet into an `extract` prompt, the extract prompt retrieves the most "important" function calls in the snippet, by whatever metric GPT deems fit
- the definitions of these function calls are located using scope-graphs from across the code-base
- the explain prompt is given snippets of these definitions

TODO: comparative analysis of cross-referencing vs no-cross-referencing.


nerdypepper:
this will need a complete rework with our recent answer-api changes (streaming, anthropic etc.), closing (but not deleting the branch) for now.

PR #175, opened 2023-02-10
tweak analytics to function on partially complete answer pipelines

nerdypepper:
in order to accomplish this, the core logic of `answer::handle` has been moved into a helper function: `_handle`. this new `handle` wraps around `_handle` and performs cleanup duties (firing analytics events) when `_handle` early exits, the call stack might look like this:

- (`handle`) receives request
- (`handle`) analytics pipeline is initialized
- (`handle -> _handle`) enter `_handle` to process request
- (`handle -> _handle`) early exit back to `handle`
- (`handle`) track analytics upto the point of exit
- (`handle`) sends response

additionally, this PR adds stage-wise-time-elapsed data to analytics (this might be done better with the tracing-timing crate).

### DIFF
--- a/server/bleep/src/webserver/answer.rs
+++ b/server/bleep/src/webserver/answer.rs
@@ -1,5 +1,6 @@
 use std::{
     collections::HashMap,
+    sync::Arc,
     time::{Duration, Instant},
 };
 
@@ -11,6 +12,7 @@ use axum::{
 };
 use futures::{Stream, StreamExt, TryStreamExt};
 use thiserror::Error;
+use tokio::sync::RwLock;
 use tracing::{debug, error, info, warn};
 use utoipa::ToSchema;
 
@@ -104,13 +106,18 @@ pub(super) async fn handle(
     Extension(app): Extension<Application>,
 ) -> Result<impl IntoResponse, (StatusCode, Json<super::Response<'static>>)> {
     // create a new analytics event for this query
-    let mut event = QueryEvent::default();
+    let event = Arc::new(RwLock::new(QueryEvent::default()));
 
     // populate analytics event
-    let response = _handle(params, app.clone(), &mut event).await;
+    let response = _handle(params, app.clone(), Arc::clone(&event)).await;
 
-    // send event to rudderstack
-    app.track_query(event);
+    if response.is_err() {
+        // send event to rudderstack
+        let event = event.read().await;
+        app.track_query(&*event);
+    } else {
+        // the analytics event is fired when the stream is consumed
+    }
 
     response
 }
@@ -118,7 +125,7 @@ pub(super) async fn handle(
 async fn _handle(
     params: Params,
     app: Application,
-    analytics_event: &mut QueryEvent,
+    event: Arc<RwLock<QueryEvent>>,
 ) -> Result<impl IntoResponse, (StatusCode, Json<super::Response<'static>>)> {
     let query_id = uuid::Uuid::new_v4();
     let mut stop_watch = StopWatch::start();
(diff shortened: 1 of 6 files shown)


server/bleep/src/webserver/answer.rs
```diff
@@ -71,20 +75,48 @@ pub struct AnswerResponse {
...
+        .stages
+        .push(Stage::new("user query", params.q.clone()));
+
     let query =
         parser::parse_nl(&params.q).map_err(|e| super::error(ErrorKind::User, e.to_string()))?;
     let target = query
         .target()
         .ok_or_else(|| super::error(ErrorKind::User, "missing search target".to_owned()))?;
 
+    let start_semantic_results = Instant::now();
```
nerdypepper, line 119: these duration variables have poorly thought out names, any suggestions?

rsdy: Can you maybe wrap these in a function instead of dumping it all in one namespace? It may be easier to see what's being measured as well, e.g

```
analytics_event.timed("semantic results", || {
... do something
})?
```

or even introduce a `struct` as state that has processing stages and internally mutates.

```
analytics_event.timed("semantic results", context.next_step);
```

nerdypepper:
@ipauler @anastasiya1155 the response now also includes the time taken by each stage of the answer-api pipeline, see sample response:

```
{
    user_id: "test_user",
    query_id: "aad4aaee-3a03-41dd-95bf-bde93682bb3f",
    stages: [
        {
            name: "user query",
            type: "string",
            data: "Where is the global SSH_KEY used?",
            time_elapsed: 35656, // time in milliseconds
        },
   ]
}
```

server/bleep/src/webserver/answer.rs
nerdypepper: the first suggest is obviously neater, but i think it requires a greater refactor of that whole module. i have opted to introduce a new timekeeping structure instead, and track the time-elapsed between the start and stop of each stage.

server/bleep/Cargo.toml
```diff
@@ -101,7 +101,7 @@ once_cell = "1.17.0"
 relative-path = "1.7.3"
 qdrant-client = { version = "0.11.6", default-features = false }
 tokenizers = "0.13.2"
-ort = { git = "https://github.com/bloopai/ort", branch = "build-rs-fix" , features = ["prefer-compile-strategy", "compile-static"] }
+ort = { git = "https://github.com/bloopai/ort", branch = "build-rs-fix"  }
```
rsdy, line 104: Don't leave this change in :)

server/bleep/src/webserver/answer.rs
```diff
@@ -71,14 +78,42 @@ pub struct AnswerResponse {
...
 
 pub async fn handle(
+    params: Query<Params>,
+    Extension(app): Extension<Application>,
+) -> Result<impl IntoResponse, impl IntoResponse> {
+    // create a new analytics event for this query
+    let mut event = QueryEvent::default();
+
+    // populate analytics event
+    let response = _handle(params, Extension(app.clone()), &mut event).await;
```
rsdy, line 88: Can you not just straight up send the `app` instance through? Is there any reason the callee needs to take `Extension<Application>`?

rsdy (changes requested):

server/bleep/src/webserver/answer.rs
rsdy: This looks a lot cleaner! 👍

server/bleep/Cargo.toml
nerdypepper: oops, thanks

server/bleep/src/webserver/answer.rs
nerdypepper: ah yes, no reason really, we may as well pass `app` here.

rsdy (approved):

nerdypepper:
in order to integrate with streaming, the analytics flow has been changed a bit:

- if the pipeline abruptly fails, the event is sent right before the handler turns over the error to the client
- if the pipeline succeeds, the event is sent after the entire stream is consumed

ggordonhall (approved):

PR #176, opened 2023-02-10
prevent re-indexing from creating duplicates in qdrant

nerdypepper:
fixes a critical bug that reduced the accuracy of `/api/answer` over time. 

previous behaviour:
- upon re-index, a new set of points are added to qdrant and the old points are not invalidated
- upsert would normally handle this, but we currently do not have a way of uniquely identifying a point 
- as every new set of points is similar (cosine-similarity) to the original set, the results of retrieval contained duplicates and thus reduced the performance of `/api/answer`.

updated behaviour:
- upon re-index, any existing points for a re-indexed file are removed from qdrant, and a new set of points for that file are calculated and inserted
- if the file is removed from the index altogether, the points corresponding to that file are removed

--- 
testing:

1. setup the repo like so:
    -  initialize a file, say `a.rs` (non-empty), and `git init && git add .`
    - measure the point count thus far: say, `N` using: `curl "http://localhost:6333/collections/documents" | jq '.result.point_count'`

2. test point addition:
    - duplicate the contents of `a.rs` into `a.rs` itself and check the change into git
    - measure the point count again, it should be `2 * N` (and not `3 * N`, which is the result without this patchset, calculated as `N (original) + 2 * N (reindexed)`)

3. test point deletion:
    - delete `a.rs` and check the change into git
    - measure the point count, it should be zero and not non-zero.

### DIFF
--- a/server/bleep/src/indexes/file.rs
+++ b/server/bleep/src/indexes/file.rs
@@ -281,12 +281,28 @@ impl Indexable for File {
 
         info!(?repo.disk_path, "file indexing finished, took {:?}", start.elapsed());
 
+        // files that are no longer tracked by the git index are to be removed
+        // from the tantivy & qdrant indices
         file_cache.retain(|k, v| {
             if v.fresh.not() {
+                // delete from tantivy
                 writer.delete_term(Term::from_field_text(
                     self.entry_disk_path,
                     &k.to_string_lossy(),
                 ));
+
+                // delete from qdrant
+                let relative_path = k.strip_prefix(&repo.disk_path);
+                if let (Some(semantic), Ok(relative_path)) = (&self.semantic, relative_path) {
+                    let semantic = semantic.clone();
+                    let reporef = reporef.to_string();
+                    let relative_path = relative_path.to_string_lossy().to_string();
+                    tokio::spawn(async move {
+                        semantic
+                            .delete_points_by_path(reporef.as_str(), relative_path.as_str())
+                            .await;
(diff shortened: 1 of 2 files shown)


server/bleep/src/indexes/file.rs
```diff
@@ -281,12 +281,28 @@ impl Indexable for File {
...
                     &k.to_string_lossy(),
                 ));
+
+                // delete from qdrant
+                let relative_path = k.strip_prefix(&repo.disk_path);
+                if let (Some(semantic), Ok(relative_path)) = (&self.semantic, relative_path) {
+                    let semantic = semantic.clone();
+                    let reporef = reporef.to_string();
+                    let relative_path = relative_path.to_string_lossy().to_string();
+                    tokio::spawn(async move {
```
rsdy, line 300: This will potentially spawn a lot of parallel deletion jobs. At this stage, I think it would be nicer to 1st create a list of paths to be deleted, then send them as one big job to qdrant and let them handle making it efficient. Would this be possible at all?

rsdy (commented):
I have one untested perf concern, otherwise it looks fine.

server/bleep/src/indexes/file.rs
nerdypepper: good point, yes this is possible, we can the filter needs some updating is all.

llogiq (approved):
Looks good to me.

rsdy (approved):

PR #180, opened 2023-02-13
Fix copy big files

anastasiya1155:
Since we use react-virtualized to render files we have issues when selecting and copying big files (missing previous lines). To fix it we can calculate the selection start and end and copy it directly from the source code.


stunningpixels:
Not working after testing, code is cutoff before the highlighted section and I get:

```

Status
·
Online
Speed of last query
·
30376ms
Report a bug
```

at the bottom of the paste

anastasiya1155:
> Not working after testing, code is cutoff before the highlighted section and I get:
> 
> ```
> 
> Status
> ·
> Online
> Speed of last query
> ·
> 30376ms
> Report a bug
> ```
> 
> at the bottom of the paste

I didn't account for cmd+A action, not it is handled properly

stunningpixels:
Shift-ENTER should go to previous match, otherwise working

PR #182, opened 2023-02-14
Remove repositories from indexes

rsdy:
Instead of `commit()`ing the changes in `delete_repo_indexes`, previously we rolled them back. The state changes were also never persisted to disk. This patch leaves the indexes in consistent state.


ggordonhall (approved):
Lgtm

PR #184, opened 2023-02-15
Use nix for CI test

rsdy:


server/bleep/src/semantic/chunk.rs
```diff
@@ -47,7 +47,7 @@ impl<'a> Chunk<'a> {
 ///
 /// # Examples
 ///
-/// ```
+/// ```no_run
```
llogiq, line 50: Why not run the test?

rsdy: It takes 120s to run this doctest.

llogiq (approved):
I guess the success is speaking for itself.

PR #192, opened 2023-02-20
shorten prompt completion by referencing code

nerdypepper:
the number of tokens produced in by the completion can be reduced by referencing to code by line numbers rather than generating code blocks.


nerdypepper:
TODO: needs to be rebased against anthropic changes.

nerdypepper:
continued in `shorten-code-gen-2` to rework against anthropic changes.

PR #197, opened 2023-02-21
tweak prompt to shorten code generation

nerdypepper:
the number of tokens produced in by the completion can be reduced by referencing to code by line numbers rather than generating code blocks.


nerdypepper:
currently on hold until https://github.com/BloopAI/bloop/pull/221 goes in.

PR #198, opened 2023-02-21
supply github user id in analytics if available

nerdypepper:


ggordonhall (approved):

server/bleep/src/webserver/answer.rs
```diff
@@ -685,6 +688,25 @@ Answer:",
     }
 }
 
+async fn github_user_id(app: &Application) -> Option<String> {
```
rsdy, line 691: Can you take a look if this exists in some form or fashion? I'd also prefer if this was closer to the other code dealing with github remotes, e.g `remotes::github::current_user(&app)` or something similar.

rsdy (commented):
I'm just seeing that this will hit GitHub on every analytics query? Is that correct?

nerdypepper:
yeah you are right, on every query to `/api/answer`, we make a query to `https://api.github.com/users/<user>`. would `github::Auth::OAuth` be a good place to store this state? another option is to make this stateless (but spoofable by clients), and calculate this info in the first response, and expect the client to send us this value on subsequent queries: `params.user_id.unwrap_or_else(|| remotes::github::user_id())`.

nerdypepper:
needs some work, closing PR without closing branch.

https://linear.app/bloopai/issue/BLO-371/bookeeping-for-the-currently-logged-in-user-if-any

PR #201, opened 2023-02-21
Break up `state.rs`

ggordonhall:
Stuck on a plane with no in-flight entertainment, so decided to make some quality-of-life improvements.

- Split out parts of `state.rs` into a new `repo.rs`
- Renaming and comments to improve readability


server/bleep/src/ctags.rs
```diff
@@ -118,7 +118,10 @@ async fn call_ctags(paths: Vec<String>, exclude_langs_list: &[&str]) -> Result<S
     Ok(String::from_utf8(output.stdout)?)
 }
 
-pub async fn get_symbols(repo_disk_path: &Path, exclude_langs: &[&str]) -> SymbolMap {
+pub async fn get_symbols_with_ctags(
```
nerdypepper, line 121: this suffix feels redundant since it already belongs to the `ctags` module, and is often invoked with a qualified path, such as `ctags::get_symbols` (as done in `benches/indexes.rs`).

ggordonhall: Yes, that's fair enough. Will make sure that they're always invoked with their qualified paths.

server/bleep/src/config.rs
```diff
@@ -66,6 +66,37 @@ pub struct Configuration {
...
+    #[clap(long)]
+    /// Key for analytics backend
+    pub analytics_key: Option<String>,
+
+    #[clap(long)]
+    /// Analytics data plane identifier
+    pub analytics_data_plane: Option<String>,
+
+    #[clap(long)]
+    /// Analytics data plane identifier
```
nerdypepper, line 90: mismatched comment, should say `Sentry DSN` or similar

server/bleep/src/webserver/repos.rs
```diff
@@ -33,7 +33,7 @@ impl From<(&RepoRef, &Repository)> for Repo {
     fn from((key, repo): (&RepoRef, &Repository)) -> Self {
         Repo {
             provider: key.backend(),
-            name: key.display_name(),
+            name: format!("{}", key),
```
nerdypepper, line 36:
```suggestion
            name: key.to_string(),
```

nerdypepper (changes requested):

server/bleep/src/webserver/repos.rs
```diff
@@ -144,7 +144,16 @@ pub(super) async fn delete_by_id(
...
+    let deleted = match app.repo_pool.get_mut(&reporef) {
+        Some(mut result) => {
+            result.value_mut().mark_removed();
+            app.write_index().queue_sync_and_index(vec![reporef]);
+            Ok(())
+        }
+        None => Err(Error::new(ErrorKind::NotFound, "Repo not found")),
+    };
+
+    Ok(deleted.map(|_| json(ReposResponse::Deleted)))
```
nerdypepper, line 156: is this new code that needs review or has this been moved from elsewhere?

nerdypepper:
the rest of the refactor looks good! i do like the new nomenclature, and `repo::RepoMetadata` feels more natural than `state::RepoHeadInfo`.

server/bleep/src/webserver/repos.rs
ggordonhall: This was a method on `RepoRef`. I've moved the logic into `webserver/repos.rs` to make consistent with other CRUD behaviour.

server/bleep/src/remotes.rs
```diff
@@ -75,6 +96,79 @@ impl From<git2::Error> for RemoteError {
...
+                    .trim_end_matches('/')
+                    .trim_end_matches(".git")
+                    .to_owned(),
+            }));
+        }
+
+        Err(())
+    }
+}
+
```
rsdy, line 171: This doesn't logically belong here. This is merely a descriptor of the repository's properties and doesn't have functional part in remote interaction. I'd prefer if these travelled with `Repository`. If you think it's too noisy in one module, submodules would work well.

server/bleep/src/remotes.rs
```diff
@@ -29,6 +32,24 @@ type GitCreds = Box<
...
+    pub address: String,
+}
+
+#[derive(Serialize, Deserialize, ToSchema, PartialEq, Eq, Clone, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum GitProtocol {
+    Https,
+    Ssh,
+}
+
```
rsdy, line 52: Same here as below.

server/bleep/src/webserver/repos.rs
rsdy: I'd prefer if this was on RepoRef, to have a definitive interface for safe deletion of a repository, which is why I originally moved it there. Now we're binding this behaviour to the API, but there's actually no reason we wouldn't want to delete a repo through other mechanisms.

In general, I'd prefer if this kind of logic moved down from the API to lower layers in a safe wrapper, than having full-blown functions in-line in API handlers.

server/bleep/src/webserver/repos.rs
```diff
@@ -33,7 +33,7 @@ impl From<(&RepoRef, &Repository)> for Repo {
     fn from((key, repo): (&RepoRef, &Repository)) -> Self {
         Repo {
             provider: key.backend(),
-            name: key.display_name(),
+            name: key.to_string(),
```
rsdy, line 36: This is not an equivalent change, look at the implementation of `display_name()`.

rsdy (changes requested):

server/bleep/src/webserver/repos.rs
ggordonhall: While I agree that this logic shouldn't live at the API layer, we also don't want it split between the API and lower layers (as it is now). I propose that we move the deletion logic up for now, and then move all the index/delete logic down to a safe wrapper in a future PR (unless you want to add to this one right now).

server/bleep/src/remotes.rs
```diff
@@ -11,10 +13,11 @@ use git2::{Cred, CredentialType, RemoteCallbacks};
 use ignore::WalkBuilder;
 use serde::{Deserialize, Serialize};
 use tracing::warn;
+use utoipa::ToSchema;
```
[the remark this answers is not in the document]
ggordonhall: This dependency was already present in `state.rs`. Agree that we want to split out but will leave for a future PR

server/bleep/src/webserver/repos.rs
ggordonhall: Good spot!

server/bleep/src/config.rs
```diff
@@ -261,6 +272,29 @@ impl Configuration {
...
+where
+    S: Serializer,
+{
+    match opt_secstr {
+        Some(secstr) => ser.serialize_some(secstr.expose_secret()),
+        None => ser.serialize_none(),
+    }
+}
+
+pub fn serialize_secret_str<S>(secstr: &SecretString, ser: S) -> Result<S::Ok, S::Error>
```
llogiq, line 288: Does this mean serialization is exfiltrating our secret strings? That sounds just a wee bit scary.

llogiq (approved):
This looks good in general, though I'm slightly worried about the deserialize impl for our secrets.

server/bleep/src/config.rs
rsdy: In practice we will want to store at least some of these secrets somewhere. I guess dumping github/git secrets may be problematic, but then where do we store them?

rsdy (approved):

nerdypepper (approved):

PR #202, opened 2023-02-22
do not unwrap on fallible canonicalize

nerdypepper:
`fs::canonicalize` can fail when attempting to follow broken symlinks, as in the case of this symlink in the tauri project: https://github.com/tauri-apps/tauri/blob/dev/core/tests/restart/LICENSE_APACHE-2.0.


llogiq (approved):
Good catch! Let's merge this!

llogiq:
Also there's still some CI problems. 🙁

nerdypepper:
for now, i can merge this without passing CI, i will take a look at broken CI in a bit.

PR #207, opened 2023-02-23
Install qdrant before starting tauri

anastasiya1155:
Changed the order of running commands, now the flow is as follows:
- `npm run start-app` or `npm run build-app`
- command is forwarded to app/desktop `tauri-dev` or `tauri-build` script
- `build-qdrant` script in run (call to ./build.sh), if successful...
- `tauri dev` or `tauri build` is executed
- Tauri CLI looks into tauri.conf.json for beforeDevCommand or beforeBuildCommand
- Vite builds JS for dev or prod usage


rsdy (approved):
Looks great!

ggordonhall (approved):
Tested and it works!

PR #209, opened 2023-02-23
Log transport errors on GitHub token validation

calyptobai:
Transport errors (`URL`, `Http`, `Json`, `JWT`, etc) are now logged and allowed during the GitHub device token validation check. Previously, any transport error like a bad network connection would result in the token getting wiped from the device. Given that the token is checked every 60 seconds, this means that if a user had a long-running tauri-based session, moving their laptop to a different room could cause their token to be deleted.

Now, only GitHub errors actually delete the token - these mean that GitHub successfully received the request and specifically returned that the token itself is not valid, for other reasons not related to network quality.

### DIFF
--- a/client/src/components/PageTemplate/index.tsx
+++ b/client/src/components/PageTemplate/index.tsx
@@ -1,6 +1,17 @@
-import React from 'react';
+import React, { useCallback, useContext, useEffect, useState } from 'react';
 import NavBar from '../NavBar';
 import StatusBar from '../StatusBar';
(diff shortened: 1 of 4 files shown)


ggordonhall (approved):

rsdy (approved):
I guess we could bubble up the error and leave it to the caller to decide what they want a layer above, but this seems fine to me, and it makes sense.

nerdypepper (approved):

PR #214, opened 2023-02-27
consume `open:` and `case:` flags in NL search

nerdypepper:


ggordonhall (approved):
Lgtm - can you add another test for the `case:` flag?

PR #222, opened 2023-03-01
do not unwrap on `strip_prefix`

nerdypepper:
it would be nice to actually get to the bottom of this error however.


nerdypepper:
@rsdy any ideas on where this `unwrap` could actually fail?

rsdy:
@nerdypepper i don't think it can, unless `repo.disk_path` is something bogus. If I understand this correctly, by definition, we're picking up files within `repo.disk_path`, so I don't think at this point this will blow up.

nerdypepper:
this happens when a file is a symlink to a path outside `repo.disk_path`. thanks @rsdy for the find!

PR #225, opened 2023-03-02
Enable all cores for the Tauri app

rsdy:
The Tauri app has been throttled, as indexing can take up significant resources. This patch tentatively removes the throttle to speed up indexing, and allows us to re-test with semantic search.


ggordonhall:
As expected, this will completely consume the CPU (clocks in at 750% usage). Currently testing impact on system performance with `max_threads / 2`.

PR #228, opened 2023-03-03
make webserver::answer::grow infallible

nerdypepper:
`grow` is can panic in 3 locations potentially:

- the index operation in `content[..snippet.start._byte]`
- the index operation in `content[snippet.end_byte..]`
- the index operation in `content[new_start..new_end]`

this PR avoids the first two by verifying bounds before indexing. the last index operation is expected to be infallible already.

### DIFF
--- a/server/bleep/src/webserver/answer.rs
+++ b/server/bleep/src/webserver/answer.rs
@@ -352,14 +352,17 @@ async fn _handle(
 
         let mut grow_size = 40;
         let grown_text = loop {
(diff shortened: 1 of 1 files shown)


nerdypepper:
the fundamental issue is still present, which is `chunk::by_tokens` is storing incorrect byte ranges into the vector DB, this PR only avoids the panic.

ggordonhall (commented):
Can you add some `error` logs when this happens? We should track the name of the repo and file that has failed.

ggordonhall (approved):

PR #230, opened 2023-03-03
Increase log density and level

ggordonhall:


llogiq (approved):
Looks good (although the _handle method on answer.rs may go away with my next PR, but don't worry).

nerdypepper (approved):

PR #231, opened 2023-03-03
Use pre-built `ort`, fix builds on Windows

calyptobai:
This PR adds support for building on Windows. Additionally, we (mostly) revert our changes in `ort`, and instead rely on pre-built binaries. On Windows, the only platform where onnxruntime has telemetry enabled, we disable it at runtime.

LFS scripts were moved to `package.json`, and the build process has been simplified slightly. It is now possible to run `npm start` to run the app.


ggordonhall (approved):
Tested and works well (on `macos-aarch64`). One question, why has `model/` been moved to `apps/desktop/src-tauri`? We also use the contents of `model/` when bleep's run without Tauri.

ggordonhall (changes requested):
Tested the Docker image and produces this error:
```
The application panicked (crashed).
Message:  could not load the library at `libonnxruntime.so`: DlOpen { desc: "libonnxruntime.so: cannot open shared object file: No such file or directory" }
Location: /usr/local/cargo/git/checkouts/ort-c6432964096301e1/d25bcf9/src/lib.rs:74

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
2023-06-20T14:57:28.565638Z DEBUG Connection{peer=Client}: h2::codec::framed_write: send frame=GoAway { error_code: NO_ERROR, last_stream_id: StreamId(0) }
2023-06-20T14:57:28.565658Z DEBUG Connection{peer=Client}: h2::proto::connection: Connection::poll; connection error error=GoAway(b"", NO_ERROR, Library)
```

server/bleep/Cargo.toml
```diff
@@ -8,8 +8,6 @@ build = "build.rs"
 [features]
 default = []
 debug = ["console-subscriber", "histogram"]
-dynamic-ort = ["ort/use-system"]
-static-ort = ["ort/use-compile", "ort/compile-static"]
```
rsdy, line 12: I'd like to keep these for flexibility to keep using our own builds with Nix.

```
[features]
default = ["dynamic-ort"]
debug = ["console-subscriber", "histogram"]
dynamic-ort = ["ort/load-dynamic"]
```

server/bleep/src/semantic.rs
```diff
@@ -193,6 +197,13 @@ impl Semantic {
                 .build()?,
         );
 
+        #[cfg(windows)]
+        {
+            unsafe {
+                ort::ort().DisableTelemetryEvents.as_ref().unwrap()(environment.env_ptr());
+            }
+        }
+
```
rsdy, line 206: I'd really like this unsafe code to move to our `ort` fork, so we can upstream it once we're happy. It is a single function, so not a huge change on either side.

server/bleep/src/semantic.rs
```diff
@@ -505,6 +516,25 @@ impl Semantic {
...
+        let lib_name = "libonnxruntime.so";
+        #[cfg(target_os = "macos")]
+        let lib_name = "libonnxruntime.dylib";
+
+        let ort_dylib_path = dylib_dir.as_ref().join(lib_name);
+
+        env::set_var("ORT_DYLIB_PATH", ort_dylib_path);
+    }
+}
+
```
rsdy, line 537: It would be good for Docker (for now) if we could just rely on having this as an environment variable, and NOT override it unconditionally. Otherwise we'll need to bump our helm templates too instead of setting it statically in the Docker build step.

This would also be helpful for local work with Nix packaging.

rsdy (changes requested):
This seems to work fine in my local env, but I'd like to ask for a few changes to be done for ease of use and separation of concerns.

Also, can we integrate the Windows qdrant build into the CI as a separate PR so we have a definitive source of builds like with the other platforms?

server/bleep/Cargo.toml
calyptobai: Does the app still work on Nix with this PR if `ort/load-dynamic` is disabled?

server/bleep/src/semantic.rs
calyptobai: Discussed offline: this applies to `--model-dir` as well, so we can take a look at this in another PR.

calyptobai:
@rsdy 

> Also, can we integrate the Windows qdrant build into the CI as a separate PR so we have a definitive source of builds like with the other platforms?

We don't build `qdrant` ourselves here, instead we simply use the [official GitHub release builds](https://github.com/qdrant/qdrant/releases).

server/bleep/src/semantic.rs
calyptobai: Swapped this with a new branch that adds `EnvBuilder::with_telemetry`

rsdy (approved):

PR #236, opened 2023-03-07
Update README & get rid of gnuisms

rsdy:
Hopefully this should make things work on older BSD toolchains.


calyptobai:
As noted on slack, seems to cause build failures due to missing `-lonnxruntime` on my end. The build script errors are resolved here now, however.

PR #242, opened 2023-03-08
Omit directories from symbol calculation

ggordonhall:
Do not try to calculate `ctags` or `scope_query` symbols on directories. Really this whole flow needs a bit of TLC.


nerdypepper (approved):

nerdypepper:
agreed, we have a similar filter for `lang_str` as well. would be nice to consolidate all of this logic.

PR #248, opened 2023-03-09
attach package metadata to analytics pipeline

nerdypepper:
package metadata in the format: `{name: "bloop", version: "0.2.3"}` wil automatically be baked into the tauri app at compile-time. note that this metadata is only attached to analytics from the desktop app currently, events from independent server analytics will contain empty package metadata.

### DIFF
--- a/apps/desktop/src-tauri/src/backend.rs
+++ b/apps/desktop/src-tauri/src/backend.rs
@@ -82,6 +82,10 @@ pub fn initialize_rudder_analytics(key: String, data_plane: String) {
             true => Some(event),
             false => None,
         })),
+        package_metadata: Some(analytics::PackageMetadata {
+            name: env!("CARGO_CRATE_NAME"),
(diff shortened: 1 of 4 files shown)


nerdypepper:
@anastasiya1155 i think the analytics panel may need an update.

apps/desktop/src-tauri/src/backend.rs
```diff
@@ -82,6 +82,10 @@ pub fn initialize_rudder_analytics(key: String, data_plane: String) {
             true => Some(event),
             false => None,
         })),
+        package_metadata: Some(analytics::PackageMetadata {
+            name: env!("CARGO_CRATE_NAME"),
+            version: env!("CARGO_PKG_VERSION"),
```
rsdy, line 87: We should probably use a git hash here instead of cargo_pkg version. https://crates.io/crates/git-version

ggordonhall (approved):

rsdy (approved):

PR #249, opened 2023-03-09
Conversational Interface

calyptobai:
**DRAFT**: This is mostly functional, there are some edge cases where the stop tokens will bleed through to the output.

---

Conversations are threaded by `user_id` currently. (As discussed offline this will be changed to `user_id` and `thread_id`, for analytics purposes).


calyptobai:
Closing in favor of #224

PR #253, opened 2023-03-13
fix: chunker works with special tokens

EdAyers:
The chunker uses the huggingface tokenizers library to make sure that chunks have the same number of tokens. However the code was panicking in the case where the tokenized encoding was producing multiple special tokens with offset `(0,0)`. This code makes the chunker more robust to this.

This is caused by padding and truncation being enabled on the tokeniser. If either of these are set then there can be more than one special tokens in the encoding, but the rust code assumes that there is exactly one special token to terminate the encoding.

todos:
- [ ] make unit tests pass
- [ ] add a check that tokenisers passed to the chunker do not have padding or truncation length set.
- [ ] There is a bug: DEDUCT_SPECIAL_TOKENS shouldn't be hardcoded to 2 because there can be multiple special tokens

### DIFF
--- a/server/bleep/src/semantic/chunk.rs
+++ b/server/bleep/src/semantic/chunk.rs
@@ -257,13 +257,13 @@ pub fn by_tokens<'s>(
     debug!("max tokens reduced to {max_tokens}");
 
     // checks whether the token at the given index contains a newline character.
-    let has_nl = |&i: &usize| {
+    let has_nl = |&i: &usize| -> bool {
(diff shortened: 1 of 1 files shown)


EdAyers:
I think this needs to be set to something else too: `pub const DEDUCT_SPECIAL_TOKENS: usize = 2`

EdAyers:
Closing this and I'll make a new PR that emits a warning if padding is turned on.

PR #255, opened 2023-03-13
Move Github repo list syncing to background

rsdy:
There's also a minor refactor here to make interfaces safer. We don't save the repository list on disk.

### DIFF
--- a/server/bleep/src/remotes.rs
+++ b/server/bleep/src/remotes.rs
@@ -201,14 +201,13 @@ pub(crate) fn gather_repo_roots(
 
 #[derive(Serialize, Deserialize, Clone, Debug)]
 pub(crate) enum BackendCredential {
-    Github(github::Auth),
+    Github(github::State),
 }
 
 impl BackendCredential {
     pub(crate) async fn validate(&self) -> Result<()> {
-        let BackendCredential::Github(auth) = self;
-
-        let client = auth.client()?;
+        let BackendCredential::Github(github) = self;
+        let client = github.client()?;
 
         match client.current().user().await {
             Ok(_) => {}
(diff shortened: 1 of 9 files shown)


server/bleep/src/remotes.rs
```diff
@@ -199,33 +200,88 @@ pub(crate) fn gather_repo_roots(
...
+        let Some(handle) = self.backends.get(&Backend::Github) else {
+	    return None;
+	};
 
-        Ok(())
+        let BackendCredential::Github(ref github) = handle.value().inner;
+        Some(github.clone())
     }
 
+    pub(crate) fn set_github(&self, gh: github::State) {
```
ggordonhall, line 255: Could the `set_` and `update_` methods be made generic?

server/bleep/src/webserver/answer.rs
```diff
@@ -286,19 +286,27 @@ async fn _handle(
     }
 
     let answer_bearer = if app.env.allow(Feature::GithubDeviceFlow) {
-        let Some(cred) = app.credentials.get(&Backend::Github) else {
+        let Some(cred) = app.credentials.github() else {
```
ggordonhall, line 289: Separate story, but it would be nice to have a way of bypassing this logic for debugging.

ggordonhall (approved):

server/bleep/src/remotes.rs
rsdy: Realistically, I don't think they can be, or at least not without larger changes. My aim here was to clean up the call sites. I don't really know how this will scale when we start adding other backends, but I'd probably leave that problem for our future selves to deal with.

nerdypepper (approved):

PR #257, opened 2023-03-14
Serve config values for the frontend

rsdy:
Currently there's only a limited number that are served. Please let me know if it looks like anything's missing.

NB: we'll probably need to rewire the Tauri app as well to use these, but a quick check would be helpful before sorting those out.

### DIFF
--- a/server/bleep/src/webserver.rs
+++ b/server/bleep/src/webserver.rs
@@ -14,6 +14,7 @@ use utoipa::ToSchema;
 mod aaa;
 pub mod answer;
 mod autocomplete;
+mod config;
 mod file;
 mod github;
 mod hoverable;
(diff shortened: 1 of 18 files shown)


anastasiya1155 (approved):

server/bleep/src/webserver.rs
```diff
@@ -41,6 +42,8 @@ pub async fn start(app: Application) -> anyhow::Result<()> {
     let bind = SocketAddr::new(app.config.host.parse()?, app.config.port);
 
     let mut api = Router::new()
+        // querying
```
ggordonhall, line 45:
```suggestion
```

ggordonhall (approved):
Lgtm!

PR #258, opened 2023-03-14
Build dependencies on macOS 11

calyptobai:
This enables backwards-compatible builds that can run on macOS 11.


rsdy (approved):

ggordonhall (commented):
Why is this necessary? Are there breaking changes between 11 and 12?

PR #261, opened 2023-03-16
Percentage indicator for individual repositories

rsdy:
It's going to be on the front-end to calculate the proper numbers, as the raw data will look like this:

```
   ["repo/reference", 0, 42]
     |                |   |=> percentage 0-100
     | reporef        |=> indexing step (0/1)
```

So a calculation as to what the progress of `repo/reference` is would be (42+69)/2 for the 2 indexing stages.

In reality, however, indexing stage 0 (repo index) will take a trivial amount of time, and we may want to discard that data completely.

You can retrieve the SSE stream through `/repos/index-status`. Note that the server will never close this stream, so if this is closed, something's definitely wrong. Realistically we could plug more events in here later, too.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -388,6 +388,7 @@ dependencies = [
  "directories",
  "dunce",
  "either",
+ "erased-serde",
  "expect-test",
  "flume",
  "futures",
@@ -1551,6 +1552,15 @@ dependencies = [
  "log",
  "regex",
  "termcolor",
+]
+
+[[package]]
+name = "erased-serde"
+version = "0.3.25"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4f2b0c2380453a92ea8b6c8e5f64ecaafccddde8ceab55ff7a8ac1029f894569"
+dependencies = [
+ "serde",
 ]
 
 [[package]]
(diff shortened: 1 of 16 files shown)


anastasiya1155:
I tried to use the new endpoint but it just times out. Also, by default, if the server does not send any events SSE connection is automatically closed, so we need some dummy events to prevent this https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format

rsdy:
@anastasiya1155 I fixed the issues with this, you'll get an output like so:

```
event:heartbeat                                                                  
                                                                                 
data:["github.com/bloopai/bloop",0,100]
data:["github.com/bloopai/bloop",1,1]
```

anastasiya1155:
I noticed that sometimes the percentage is not linear, so it goes to 1, 28, and then back to 1, 0, and up again. It can happen a couple of times before the repo is fully synced. Not sure if that's important though, fyi @stunningpixels

ggordonhall:
@anastasiya1155 I have noticed this too. When syncing bloop the percentage repeatedly drops back down 5% (from about 30%), before suddenly jumping up to completion.

PR #262, opened 2023-03-17
Conversational Interface

ggordonhall:
This adds support for a chat interface. With this bloop supports follow up queries which reference previous questions and answers in the conversational history. These are rendered in a chat window on the right-hand-side of the search results.

This bundles a host of changes, including:
- [Rake](https://docs.rs/rake/latest/rake/) keyword extraction for search queries
- Change of provider to `OpenAI`
- Rewritten prompts which support `OpenAI`'s 'message' based prompt-format

The abstractions in the `answer.rs` module will be re-worked in a future PR.


anastasiya1155 (approved):

server/bleep/src/webserver/answer.rs
```diff
@@ -490,9 +678,18 @@ fn grow(doc: &ContentDocument, snippet: &Snippet, size: usize) -> Option<String>
     Some(content[new_start_byte..new_end_byte].to_owned())
 }
 
+fn get_keywords(query: &str) -> String {
+    let stop_words_path = "../../../server/stopwords.txt";
```
nerdypepper, line 682: this path will not always resolve. i propose that we use `include_str!` here and bake stopwords into the binary (just 3.6Kb). 

and then calculate `sw` in the next line as:

```
let stop_words: HashSet<String> = sw.lines().map(ToOwned::to_owned).collect();
let sw = stop_words.into();
```

server/bleep/src/webserver/answer.rs
```diff
@@ -309,151 +399,250 @@ async fn _handle(
...
-        let mut grow_size = 40;
-        let grown_text = loop {
-            if let Some(grown_text) = grow(&doc, relevant_snippet, grow_size) {
-                let token_count = semantic.gpt2_token_count(&grown_text);
-                info!(%grow_size, %token_count, "growing ...");
-                if token_count > 2000 || grow_size > 100 {
-                    break grown_text;
+        .push(Stage::new("start", &query).with_time(stop_watch.lap()));
+
+    loop {
```
rsdy, line 431: This is an unbounded `loop` inside a very long function, which make this logic inherently hard to reason about. It would be good to either make this shorter, or change this to `loop` to something with explicit bounds.

rsdy (commented):
I think there are serious code improvements that should be made before we can review this in earnest, otherwise as an outsider this will be very hard to pick up.

calyptobai (approved):
Refactoring comments noted in: https://linear.app/bloopai/issue/BLO-406/clean-up-answer-endpoint

PR #265, opened 2023-03-17
fix: emit an error if chunker is called with a tokenizer with padding

EdAyers:
The chunker can panic if called with a tokenizer with padding or truncation enabled. This adds an error message so it is easier to catch this issue in the future.

Previous issue: https://github.com/BloopAI/bloop/pull/253

### DIFF
--- a/server/bleep/src/semantic/chunk.rs
+++ b/server/bleep/src/semantic/chunk.rs
@@ -213,6 +213,11 @@ pub fn by_tokens<'s>(
     max_lines: usize,
     strategy: OverlapStrategy,
 ) -> Vec<Chunk<'s>> {
+    if tokenizer.get_padding().is_some() || tokenizer.get_truncation().is_some() {
+        error!(
+            "This code can panic if padding and truncation are not turned off. Please make sure padding is off."
+        );
+    }
     let min_tokens = token_bounds.start;
     // no need to even tokenize files too small to contain our min number of tokens
     if src.len() < min_tokens {


EdAyers:
I have checked that bleep doesn't trigger the error by running `cargo run -p bleep --release -- --source-dir ~/my_example_repo` and then  `curl -v "localhost:7878/api/repos/indexed"` tells me that the repository got indexed successfully but I do not observe any errors in the terminal.

EdAyers:
ok this is RTM now. I ran tests locally, including with changing the `error!` to a `panic!` and the tests still passed. However I don't really understand rust so it would be great if someone could verify that the error isn't being triggered in normal operation of bleep.

nerdypepper:
if i understand correctly, those two settings when enabled on a tokenizer, can crash the app during chunking. why don't we disable it (if possible, this change would go inside [here](https://github.com/BloopAI/bloop/blob/66fadccef279970743ea2e851a7ecbf78ddecd99/server/bleep/src/semantic.rs#L69)) or upload a tokenizer to git-lfs that does not use these two settings?

EdAyers:
I think it is disabled in [tokenizer.json](https://github.com/BloopAI/bloop/blob/66fadccef279970743ea2e851a7ecbf78ddecd99/model/tokenizer.json) ('truncation' and 'padding' arguments are set to null). The problem came when using a the `chunk.rs` file directly in the search-evaluation repository, and it is not obvious that the panic was because I had padding and truncation turned on, so this error just makes it clear to users that they should be off.

nerdypepper:
i think on the off-chance that the tokenizer is updated in bloop, this error could come in handy.

PR #270, opened 2023-03-19
Prevent form submission during Japanese input composition

ainoya:
Made improvements to the handling of Japanese input by using the onCompositionStart and onCompositionEnd events to prevent form submission during the composition phase. Specifically, I added a composing state and modified the handleEnter function to check this state before allowing form submission. This ensures that the form is not submitted until the user has finished composing their input.

reference: [Element: compositionstart event \- Web APIs \| MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionstart_event)


anastasiya1155 (approved):

anastasiya1155:
@ainoya can you please rebase onto the latest main? It has an eslint fix in another place

ainoya:
@anastasiya1155 okay, I have confirmed that there are no lint errors after performing a rebase.

PR #277, opened 2023-03-20
add more detail to analytics events

nerdypepper:
- each answer progress stage contains information relevant to that stage, the rephrase stage for example, contains the finally rephrased query, instead of the query to be rephrased.
- similarly, the select and explain stages now contain the entire prompt used, along with a thread of the conversation so far.
- minor fix: `grow` contained a duplicate bounds check, possibly an artifact of a rebase/merge.


ggordonhall (approved):
Lgtm

PR #299, opened 2023-03-21
Support macOS 11

rsdy:
This PR swaps out the `qdrant` binary to a backwards compatible one. Would be great if we could test if we need to pull in a new `ctags` as well.


ggordonhall (approved):
This looks ok to me. @calyptobai can you test this and make sure that it works on your machine?

PR #312, opened 2023-03-22
add missing analytics stage

nerdypepper:
- the "semantic_results" stage previously contained filtered semantic results
- this patch adds back the original analytics stage prior to filtering: "semantic_results", and renames the existing stage to "filtered_semantic_results"

### DIFF
--- a/server/bleep/src/webserver/answer.rs
+++ b/server/bleep/src/webserver/answer.rs
@@ -242,6 +242,11 @@ async fn search_snippets(
             }
         })
         .collect();
+
+    Ok(all_snippets)
+}
+
+fn deduplicate_snippets(all_snippets: Vec<Snippet>) -> Vec<Snippet> {
     let mut snippets = Vec::new();
     let mut chunk_ranges_by_file: HashMap<String, Vec<std::ops::Range<usize>>> = HashMap::new();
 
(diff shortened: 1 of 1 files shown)


server/bleep/src/webserver/answer.rs
```diff
@@ -429,10 +434,16 @@ async fn handle_inner(
...
+                info!("Retrieved {} snippets", all_snippets.len());
+
+                event.write().await.stages.push(
+                    Stage::new("semantic_results", &all_snippets).with_time(stop_watch.lap()),
+                );
+
+                let filtered_snippets = deduplicate_snippets(all_snippets);
 
-                if s.is_empty() {
+                if filtered_snippets.is_empty() {
```
ggordonhall, line 446: What happens if `all_snippets` is empty? Will this trigger an error in `deduplicate_snippets`?

ggordonhall (approved):

ggordonhall (approved):

server/bleep/src/webserver/answer.rs
nerdypepper: `deduplicate_snippets` is expected to only remove duplicates, if an empty vector is passed in, and empty vector is returned.

PR #314, opened 2023-03-22
implement snippet deduplication using MMR

nerdypepper:
- deduplication is on par with `main`
- `main` refuses to collect more than 4 snippets from the same file, MMR does not have an artificial limit.

### DIFF
--- a/server/bleep/src/webserver/answer.rs
+++ b/server/bleep/src/webserver/answer.rs
@@ -1,5 +1,4 @@
 use std::{
-    collections::HashMap,
     sync::Arc,
     time::{Duration, Instant},
 };
@@ -11,6 +10,7 @@ use axum::{
     Extension,
 };
 use futures::{stream, Stream, StreamExt, TryStreamExt};
+use qdrant_client::qdrant::{vectors, ScoredPoint};
 use secrecy::ExposeSecret;
 use thiserror::Error;
 use tokio::sync::RwLock;
@@ -24,7 +24,7 @@ use crate::{
     query::parser,
     remotes,
     repo::RepoRef,
-    semantic::Semantic,
+    semantic::{self, Semantic},
     Application,
 };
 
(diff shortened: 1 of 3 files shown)


server/bleep/src/webserver/answer.rs
```diff
@@ -1092,3 +1085,52 @@ impl StopWatch {
         duration
     }
 }
+
+fn dot(a: &[f32], b: &[f32]) -> f32 {
```
ggordonhall, line 1089: Can you move these functions into `semantic.rs`. They are not specific to the `answer` pipeline and could be applied to any set of semantic search results.

server/bleep/src/webserver/answer.rs
```diff
@@ -82,6 +82,9 @@ pub struct Snippet {
     pub start_byte: usize,
     pub end_byte: usize,
     pub score: f32,
+
+    #[serde(skip)]
+    pub vector: Vec<f32>,
```
ggordonhall, line 87: Can we rename this `embedding` and add a comment explaining where it is used?

server/bleep/src/webserver/answer.rs
```diff
@@ -234,49 +248,23 @@ async fn search_snippets(
...
-                    .or_insert_with(Vec::new)
-                    .push(std::ops::Range {
-                        start: snippet.start_line,
-                        end: snippet.end_line,
-                    });
-                snippets.push(snippet);
-            }
-        }
+fn deduplicate_snippets(all_snippets: Vec<Snippet>, query_embedding: Vec<f32>) -> Vec<Snippet> {
+    let lambda = 0.5;
```
ggordonhall, line 260: Would be nice to have a comment explaining what `lambda` does.

server/bleep/src/webserver/answer.rs
```diff
@@ -442,7 +438,16 @@ async fn handle_inner(
...
                 );
 
-                let filtered_snippets = deduplicate_snippets(all_snippets);
+                let query_embedding = semantic.embed(rephrased_query).map_err(|e| {
+                    error!("failed to embed query: {}", e);
+                    Error::internal(e)
+                })?;
+                let filtered_snippets = deduplicate_snippets(all_snippets, query_embedding);
+
+                event.write().await.stages.push(
```
nerdypepper, line 447: moved the analytics event up here, so we can see if `filtered_snippets` is empty in the pipeline.

ggordonhall (approved):

PR #319, opened 2023-03-22
Update internal product name for more consistent directory use

rsdy:


ggordonhall (approved):
After this change, how many caches will we have and what will they be called?

anastasiya1155:
We still have two folders in the cache 
Screenshot 2023-03-23 at 08 36 55

For some reason Tauri creates 'ai.bloop' instead of using the server folder. But I noticed that other apps have the same thing, I have a couple of other folders that have three words and four words separated by a dot, where the first three words are the same, like 'co.teamport.around' and 'co.teamport.around.ShipIt'

Issue #289, opened 2023-03-21
3.1  is not work in ubuntu 20.04 


felixduan0:
**Describe the bug**
try both deb and appimage
return this 
```
$: ./bloop_0.3.1_amd64.AppImage 
bloop: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by bloop)
bloop: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by bloop)
bloop: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by bloop)
bloop: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by bloop)
bloop: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.35' not found (required by /tmp/.mount_bloop_DRx5Zm/usr/lib/libwebkit2gtk-4.0.so.37)
bloop: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /tmp/.mount_bloop_DRx5Zm/usr/lib/libwebkit2gtk-4.0.so.37)
bloop: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /tmp/.mount_bloop_DRx5Zm/usr/lib/libwebkit2gtk-4.0.so.37)
bloop: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /tmp/.mount_bloop_DRx5Zm/usr/lib/libwebkit2gtk-4.0.so.37)
bloop: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by /tmp/.mount_bloop_DRx5Zm/usr/lib/libwebkit2gtk-4.0.so.37)
bloop: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /tmp/.mount_bloop_DRx5Zm/usr/lib/libgtk-3.so.0)
bloop: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /tmp/.mount_bloop_DRx5Zm/usr/lib/libgdk-3.so.0)
[...]

PR #337, opened 2023-03-24
Build tauri releases on ubuntu-20.04

rsdy:


rsdy:
Seems like due to our dependencies we can't build on older platforms.

#289 left unresolved. If anybody is willing to pursue this, I'm happy to merge a patch that's compatible.

PR #341, opened 2023-03-27
attach device-id in sentry

nerdypepper:


ggordonhall (approved):
Looks good, changed the base to `main`

PR #344, opened 2023-03-27
Remove dashmap

rsdy:
The DashMap API is a footgun for any async application. The core of the issue is that the external references (including inside iterators) allow inadvertently carrying locks across `async` boundaries, which results in a deadlock.

The `scc` crate fixes this issue by explictily disallowing unsafe use of the API , eliminating this entire bug-class.

To be extra cautious, I removed every use of `dashmap`, and now the granular multi-threaded locks work as expected.


calyptobai (approved):
LGTM :+1:

Our heavy reliance on sharing hashmaps suggests deeper architectural issues here.

nerdypepper (approved):

PR #364, opened 2023-04-03
Add new onboarding step for adding public repos

anastasiya1155:


stunningpixels:
Thanks, overall very promising. A few points of feedback:

Can we skip the 'verify step' ? Can be a bit confusing, I thought at first that it had been added to be synced but later found out it hadn't.

We should also have the ability (like with any list) to remove repos from the list.

Also getting this error
Screenshot 2023-04-04 at 14 26 17

Also think we need to rework the add repos flow to present these options clearer:
Screenshot 2023-04-04 at 14 29 48

Screenshot 2023-04-04 at 14 30 07

ggordonhall:
Is this option also available for cloud deployments?

anastasiya1155:
> Is this option also available for cloud deployments?

The onboarding flow is different, but in settings there is an option to add public repo as well as private github repos from organisation

PR #365, opened 2023-04-03
Index git branches

rsdy:
This is the indexing side of the story.

Since we don't yet do bare checkouts, we can run ctags on the `git` checkout of the repository.

It also seems like calculating the "most used language" across a repository can be trivially moved into inside file indexing loop with, and lazily populated.

This would potentially get us some performance gains, as we only have to read up the whole repo once (instead of twice currently). However, doing this should also depend on our future plans for calculating the `symbol_locations` table, what to do with `ctags`, etc.

The branches will be stored in the databases in the following format.

```
["head", "main", "origin/HEAD", "origin/add-github-username-rudder", "origin/add-user-to-sentry", "origin/ai-code-nav", "origin/auth-bastion", "origin/batch-embed", "origin/bloop-cursor",  "v0.0.1", "v0.1.0", "v0.2.0", "v0.2.1", "v0.2.2", "v0.2.3", "v0.3.0", "v0.3.1", "v0.3.2", "v0.3.3"]
```

### DIFF
--- a/server/bleep/src/repo/iterator/git.rs
+++ b/server/bleep/src/repo/iterator/git.rs
@@ -9,21 +9,23 @@ use std::{collections::HashMap, path::Path};
 
 pub enum BranchFilter {
     All,
+    Head,
     Select(RegexSet),
 }
 
 impl BranchFilter {
-    fn filter(&self, branch: &str) -> bool {
+    fn filter(&self, is_head: bool, branch: &str) -> bool {
         match self {
             BranchFilter::All => true,
             BranchFilter::Select(patterns) => patterns.is_match(branch),
+            BranchFilter::Head => is_head,
         }
     }
 }
 
 impl Default for BranchFilter {
     fn default() -> Self {
-        Self::All
+        Self::Head
     }
 }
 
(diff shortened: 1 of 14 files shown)


server/bleep/src/ctags.rs
```diff
@@ -24,6 +24,30 @@ use tracing::{debug, warn};
 pub static CTAGS_BINARY: OnceCell<PathBuf> = OnceCell::new();
 pub type SymbolMap = HashMap<PathBuf, Vec<Symbol>>;
 
+// Extract symbols using Ctags for all languages which are not covered by a more
```
nerdypepper, line 27: as the comment suggests, this might be created dynamically in the future, const-ness might not be an option then. alright to keep for now though i think.

nerdypepper (commented):
will have to do a round of testing as well.

server/bleep/src/repo/iterator/git.rs
```diff
@@ -0,0 +1,151 @@
...
+        match self {
+            BranchFilter::All => true,
+            BranchFilter::Select(patterns) => patterns.is_match(branch),
+        }
+    }
+}
+
+impl Default for BranchFilter {
+    fn default() -> Self {
+        Self::All
```
ggordonhall, line 26: Does this mean that by default we'll index _all_ branches. Shouldn't we only index `HEAD` by default?

ggordonhall (approved):

nerdypepper (approved):
pr looks good & works. since branches are kind of an implementation detail, i think we should turn the `branch` filter on regex search into a no-op until the UI is built around it (it already is a no-op on NL searches). thoughts?

PR #370, opened 2023-04-04
Use GitHub username or id to identify

rsdy:
This adds a persistent tracking_id that's user-aware. The core of the logic is in `webserver/config.rs`, and would be good to get another pair of eyes if this is what we want.

This also opens the opporunity to potentially reset this in the future on a per-user basis, but that's unimplemented currently.

### DIFF
--- a/server/bleep/src/lib.rs
+++ b/server/bleep/src/lib.rs
@@ -22,6 +22,7 @@ use console_subscriber as _;
 #[cfg(target_os = "windows")]
 use dunce::canonicalize;
 use scc::hash_map::Entry;
+use state::PersistedState;
 #[cfg(not(target_os = "windows"))]
 use std::fs::canonicalize;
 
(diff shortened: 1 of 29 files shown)


server/bleep/src/state.rs
```diff
@@ -47,6 +48,95 @@ pub struct StateSource {
...
+
+impl ToString for ApplicationSeed {
+    fn to_string(&self) -> String {
+        self.0.to_string()
+    }
+}
+
+impl Default for ApplicationSeed {
+    fn default() -> Self {
+        let seed: [u8; 32] = rand::random();
```
ggordonhall, line 115: Why use this rather than `uuid`?

server/bleep/src/state.rs
```diff
@@ -47,6 +48,95 @@ pub struct StateSource {
     cookie_key: Option<PathBuf>,
 }
 
+pub struct PersistedState<T> {
```
ggordonhall, line 51: Can you add a comment explaining what this struct does?

server/bleep/src/lib.rs
```diff
@@ -129,10 +158,14 @@ impl Application {
         Ok(Self {
             indexes: Arc::new(Indexes::new(config.clone(), semantic.clone())?),
             background: BackgroundExecutor::start(config.clone()),
+            prior_conversational_store: Arc::default(),
             repo_pool: config.source.initialize_pool()?,
             cookie_key: config.source.initialize_cookie_key()?,
             credentials: config.source.initialize_credentials()?.into(),
-            prior_conversational_store: Arc::default(),
+            user_store: config.source.load_or_default("user_store")?,
+            tracking_seed: config
```
ggordonhall, line 166: Just to be clear, how does this work with cloud deployments where there is more than 1 user per `Application`? What is the `tracking_seed` in this case?

ggordonhall (commented):
Will we need to make some changes to `answer.rs` and `analytics.rs` to propagate the user identifier into the analytics DB? At the moment the frontend passes a `user_id` to `/answer`.

@anastasiya1155 I'd like to get your input on this. What is the `user_id` passed to `/answer`?

anastasiya1155:
> Will we need to make some changes to `answer.rs` and `analytics.rs` to propagate the user identifier into the analytics DB? At the moment in the frontend passes a `user_id` to `/answer`.
> 
> @anastasiya1155 I'd like to get your input on this. What is the `user_id` passed to `/answer`?

We use unique device id as a user identifier at the moment

server/bleep/src/lib.rs
rsdy: In cloud deployments, the `Application`-level tracking seed can be used to identify the instance. We don't expose this currently, but we can in the future.

If there are multiple users in the system, we will always use the per-user values in `user_store`.

server/bleep/src/state.rs
rsdy: There's no good reason, I didn't think of `uuid` at all. I can change it if that seems more consistent.

nerdypepper (approved):
this also opens up the gates for adding gh user-ids in analytics events!

PR #371, opened 2023-04-05
Github repos transitioning into local ones

rsdy:
The issue here seems to be that we only persist the state sometimes during the processing pipeline, which allows us losing information: If the user quits the app after the initial git sync, but before indexing completes, there's a chance we will not have saved the repo status.

Adding explicit persistence to these synchronization points should fix this.


ggordonhall (approved):
Lgtm

nerdypepper:
nice find.

PR #376, opened 2023-04-13
turn NLQuery filters into exact matches

nerdypepper:
in the future, it would be nice to support full blown regex here, that way an exact match would be achieved by writing a filter of the form repo:/^string$/ and a substring match would be achieved by writing a filter of the form repo:string.

### DIFF
--- a/server/bleep/src/semantic.rs
+++ b/server/bleep/src/semantic.rs
@@ -326,7 +326,7 @@ fn make_kv_filter(key: &str, value: &str) -> FieldCondition {
     FieldCondition {
         key,
         r#match: Some(Match {
-            match_value: MatchValue::Text(value).into(),
+            match_value: MatchValue::Keyword(value).into(),
         }),
         ..Default::default()
     }


server/bleep/src/semantic.rs
```diff
@@ -332,6 +332,18 @@ fn make_kv_filter(key: &str, value: &str) -> FieldCondition {
     }
 }
 
+fn make_kv_filter_strict(key: &str, value: &str) -> FieldCondition {
```
ggordonhall, line 335: Should this also be used in the `delete_points_by_path` function?

ggordonhall (approved):

server/bleep/src/semantic.rs
nerdypepper: yes it should. made this a separate utility in order to avoid breaking existing behaviour. will swap it out and give this a test.

nerdypepper: works as expected from a quick test:
- index a repo
- count points in qdrant
- delete tantivy index
- re-index the repo
- count points in qdrant again

the count should not double, it should stay the same.

rsdy (approved):

ggordonhall (approved):

PR #381, opened 2023-04-13
Rudderstack cleanup 3

rsdy:
Moves the `identify` call to the backend.


server/bleep/src/analytics.rs
```diff
@@ -123,11 +133,39 @@ impl RudderHub {
         }
     }
 
-    pub fn track_query(&self, user: &crate::webserver::middleware::User, event: QueryEvent) {
+    pub fn identify(&self, self_serve: bool, org_name: Option<&str>, remote_login: &str) {
```
ggordonhall, line 136: We should rename `self_serve` to `cloud`. Also, `remote_login` is a little confusing, since this looks like it's the remote host username - could have `remote_username` instead.

ggordonhall (approved):
Backend changes look good. @anastasiya1155 does anything else need to be modified on the FE?

server/bleep/src/analytics.rs
```diff
@@ -123,11 +133,39 @@ impl RudderHub {
...
+    pub fn identify(&self, cloud: bool, org_name: Option<&str>, github_username: &str) {
+        let event = Identify {
+            user_id: Some(self.tracking_id(&User(Some(github_username.to_string())))),
+            anonymous_id: None,
+            traits: Some(json!( {
+            "isSelfServe": cloud,
+            "githubUsername": github_username,
+            "orgName": org_name,
+            "deviceId": self.device_id(),
+            })),
```
nerdypepper, line 145:
```suggestion
            traits: Some(json!({
                "isSelfServe": cloud,
                "githubUsername": github_username,
                "orgName": org_name,
                "deviceId": self.device_id(),
            })),
```

nerdypepper: the old behaviour here in the FE was to use `org_name` or the empty string if absent, the current behaviour is to use `org_name` or null if absent. does the change in behaviour matter here @anastasiya1155?

another change i might add is to set `deviceId` to `self.device_id().trim()`.

anastasiya1155 (approved):

PR #394, opened 2023-04-18
prepend repo queries with `github.com/` when necessary

nerdypepper:
- remote repo names are prepended with `github.com/`
- this changeset also removes repo or lang filters if they are empty

### DIFF
--- a/server/bleep/src/semantic.rs
+++ b/server/bleep/src/semantic.rs
@@ -178,33 +178,50 @@ impl Semantic {
             anyhow::bail!("no search target for query");
         };
 
-        let repo_filter: qdrant_client::qdrant::Condition = {
+        let repo_filter = {
             let conditions = parsed_query
                 .repos()
-                .map(|r| make_kv_filter("repo_name", r).into())
-                .collect();
+                .map(|r| {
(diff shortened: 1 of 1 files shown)


nerdypepper:
this is a stopgap solution, ideally the index will contain `repo.display_name()` for searching.

server/bleep/src/semantic.rs
```diff
@@ -178,33 +178,50 @@ impl Semantic {
...
 
-        let repo_filter: qdrant_client::qdrant::Condition = {
+        let repo_filter = {
             let conditions = parsed_query
                 .repos()
-                .map(|r| make_kv_filter("repo_name", r).into())
-                .collect();
+                .map(|r| {
+                    if r.contains('/') {
+                        format!("github.com/{r}")
```
rsdy, line 186: I would make this conditional, so we never end up with `github.com/github.com/org/repo` accidentally?

ggordonhall (approved):
Tested for search and deletion. 

We should make sure to bundle a more robust solution with other breaking changes to the index.

rsdy (approved):

PR #395, opened 2023-04-18
Make sure the index folder exists

rsdy:
I'm not sure how this regression creeped in, but in my tests this fixes it.

### DIFF
--- a/server/bleep/src/state.rs
+++ b/server/bleep/src/state.rs
@@ -96,7 +96,7 @@ impl<T> Clone for PersistedState<T> {
 
 impl StateSource {
     pub(crate) fn set_default_dir(&mut self, dir: &Path) {
-        std::fs::create_dir_all(&dir).unwrap();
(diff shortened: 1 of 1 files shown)


server/bleep/src/state.rs
```diff
@@ -96,6 +96,8 @@ impl<T> Clone for PersistedState<T> {
 
 impl StateSource {
     pub(crate) fn set_default_dir(&mut self, dir: &Path) {
+        std::fs::create_dir_all(&dir).unwrap();
```
ggordonhall, line 99: We should add an informative error message here, printing `dir` in the event that this errors.

ggordonhall (approved):

rsdy:
@nerdypepper I think it's a terminology confusion. The root index dir is not the same as the tantivy index dirs (which may be a few layers down from the root). It would be good to come up with a better, less confusing terminology for these, since we also have qdrant and potentially other intermediate indexes, too.

nerdypepper (approved):

PR #404, opened 2023-04-21
Unify NL and grep endpoints

rsdy:
WIP, but early feedback on the structure would be helpful.
It would be helpful in particular to get some feedback on calculating the `ResultStats`  for the semantic results.


server/bleep/src/semantic/execute.rs
```diff
@@ -0,0 +1,68 @@
...
+        .map(|((relative_path, repo_name, repo_ref, lang), snippets)| {
+            QueryResult::Snippets(crate::snippet::SnippedFile {
+                relative_path,
+                repo_name,
+                repo_ref,
+                snippets,
+                lang,
+            })
+        })
+        .collect::<Vec<_>>();
```
ggordonhall, line 60: We should add MMR deduplication. This will probably break the paging strategy, but we can implement the same approach that we take for Tantivy searches.

server/bleep/src/webserver/semantic.rs
```diff
@@ -1,84 +1,40 @@
...
-
-#[derive(Serialize)]
-pub(super) struct SemanticResponse {
-    chunks: Vec<serde_json::Value>,
-}
-
-impl super::ApiResponse for SemanticResponse {}
-
 /// Get details of an indexed repository based on their id
 //
```
ggordonhall, line 12: I think we can remove this comment too

server/bleep/src/webserver/semantic.rs
```diff
@@ -1,84 +1,40 @@
...
-#[utoipa::path(get, path = "/repos/indexed/:ref",
-    responses(
-        (status = 200, description = "Execute query successfully", body = SemanticResponse),
-        (status = 400, description = "Bad request", body = EndpointError),
-        (status = 500, description = "Server error", body = EndpointError),
-    ),
-)]
-pub(super) async fn raw_chunks(
-    Query(args): Query<Args>,
+pub(super) async fn complex_search(
```
ggordonhall, line 13: Why do we need this? Shouldn't this just be the behaviour that `/q` exposes?

ggordonhall (commented):
Coming along nicely! Left a couple of thoughts

server/bleep/src/webserver/semantic.rs
rsdy: I'd want to transition this to `/q` after it stabilizes, but it's useful to have them compared side-by-side until that happens.

server/bleep/src/query/parser.rs
```diff
@@ -411,13 +442,14 @@ pub fn parse(query: &str) -> Result<Vec<Query<'_>>, ParseError> {
...
 
-pub fn parse_nl(query: &str) -> Result<NLQuery<'_>, ParseError> {
+pub fn parse_nl(query: &str) -> Result<ParsedQuery<'_>, ParseError> {
     let pairs = PestParser::parse(Rule::nl_query, query).map_err(Box::new)?;
 
     let mut repos = HashSet::new();
     let mut langs = HashSet::new();
     let mut branch = HashSet::new();
     let mut target: Option<Literal> = None;
+    let mut force_parsing_as = ForceParsingAs::Semantic;
```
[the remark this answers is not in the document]
rsdy: I wouldn't try to do this in this PR, as this may have quite an impact on the very light-weight parsing code we currently use for NL queries.

ggordonhall:
Minor thing. Since we're removing `utoipia` here, we can also remove this workflow: https://github.com/BloopAI/bloop/blob/main/.github/workflows/documentation.yml and the `bloop/scripts` dir, which is only used by that workflow.

nerdypepper (approved):
lgtm. can we clean up the commit history into:

- combine semantic & regex endpoints
- remove utoipa

and then subsequently merge without squashing?

server/bleep/src/semantic/execute.rs
```diff
@@ -0,0 +1,71 @@
...
+    Ok(QueryResponse {
+        count: data.len(),
+        metadata: PagingMetadata::new(params.page, params.page_size, None),
+        stats: ResultStats::default(),
```
nerdypepper, line 68: the `page_size` value here is bound to be invalid because of deduplication. but i don't think this is being used anywhere currently. we can't fix paging metadata without being inefficient here.

server/bleep/src/webserver/autocomplete.rs
```diff
@@ -7,27 +7,16 @@ use crate::{
...
-#[utoipa::path(
-    get,
-    path = "/autocomplete",
-    params(ApiQuery),
-    responses(
-        (status = 200, description = "Execute query successfully", body = Response),
-        (status = 400, description = "Bad request", body = EndpointError),
-        (status = 500, description = "Server error", body = EndpointError),
-    ),
-)]
```
nerdypepper, line 30: this hunk should be moved into utiopa-removal PR.

server/bleep/src/webserver.rs
```diff
@@ -261,53 +256,6 @@ impl<'a> From<EndpointError<'a>> for Response<'a> {
     }
 }
 
-#[derive(OpenApi)]
```
nerdypepper, line 264: this hunk should be moved.

server/bleep/src/webserver.rs
```diff
@@ -91,10 +89,7 @@ pub async fn start(app: Application) -> anyhow::Result<()> {
         api = middleware::local_user(api, app.clone());
     }
 
-    api = api
```
nerdypepper, line 94: this hunk should be moved.

server/bleep/src/webserver.rs
```diff
@@ -7,8 +7,6 @@ use tower::Service;
 use tower_http::services::{ServeDir, ServeFile};
 use tower_http::{catch_panic::CatchPanicLayer, cors::CorsLayer};
 use tracing::info;
-use utoipa::OpenApi;
```
nerdypepper, line 10: this hunk should be moved.

server/bleep/src/webserver.rs
```diff
@@ -221,7 +216,7 @@ struct EndpointError<'a> {
 
 /// The kind of an error
 #[allow(unused)]
-#[derive(serde::Serialize, PartialEq, Eq, ToSchema, Debug)]
```
nerdypepper, line 224: this hunk should be moved.

nerdypepper (changes requested):
the rest of the patch looks good.

nerdypepper (approved):

server/bleep/src/webserver/answer.rs
```diff
@@ -184,9 +166,12 @@ pub(super) async fn handle(
 }
 
 fn parse_query(query: &str) -> Result<String, Error> {
-    Ok(parser::parse_nl(query)
-        .map_err(Error::user)?
-        .target()
+    let ParsedQuery::Semantic(q): ParsedQuery = parser::parse_nl(query)
+        .map_err(Error::user)? else {
+	    return Err(Error::new(ErrorKind::User, "can't do that here"));
```
ggordonhall, line 171: Can you add a more informative error message?

server/bleep/src/semantic.rs
```diff
@@ -473,3 +588,34 @@ pub fn deduplicate_with_mmr(
     }
     idxs
 }
+
+pub fn deduplicate_snippets(
+    mut all_snippets: Vec<Payload>,
+    query_embedding: Embedding,
+) -> Vec<Payload> {
+    const SNIPPET_COUNT: usize = 20;
```
ggordonhall, line 596: `k` should be passed as a parameter. We don't need to set `SNIPPET_COUNT` here.

ggordonhall (approved):

server/bleep/src/semantic.rs
```diff
@@ -592,12 +592,11 @@ pub fn deduplicate_with_mmr(
 pub fn deduplicate_snippets(
     mut all_snippets: Vec<Payload>,
     query_embedding: Embedding,
+    output_count: usize,
```
ggordonhall, line 595: We can just call this `k` and remove the assignment

rsdy: No, I'd like this to be a human-readable variable name at least in high-level APIs.

ggordonhall: `k` is a conventional name in search. It's common to talk about the `top-k` elements in a ranked list for example.

PR #412, opened 2023-04-24
Add line filters for file endpoint

rsdy:


anastasiya1155 (approved):

server/bleep/src/webserver/file.rs
```diff
@@ -2,41 +2,148 @@ use std::sync::Arc;
...
+        assert_eq!(
+            split_by_lines(
+                text,
+                &indices,
+                &Params {
+                    line_start: None,
+                    line_end: None
+                }
+            )
+            .unwrap_or_else(|_| panic!("bad")),
```
nerdypepper, line 93: this could be `.expect("bad")`.

nerdypepper (approved):
lgtm

server/bleep/src/webserver/file.rs
rsdy: Trying to get around the type in the`Err` variant not being `Debug`, so I don't think `expect` will work.

ggordonhall (approved):

PR #427, opened 2023-04-27
Fix `case:ignore` in path searches

calyptobai:
Now we make sure to set the `case_insensitive` option on the generated regex.


nerdypepper (approved):
we have the same regression in `RepoReader`.

ggordonhall (approved):

PR #432, opened 2023-04-27
Remove git2 in favour of gitoxide

rsdy:


ggordonhall (approved):
This looks good. Much cleaner, and lots of nice new error types.

I have tested this on desktop, but not with GitHub App credentials.

PR #448, opened 2023-05-03
introduce analytics to tooling

nerdypepper:


nerdypepper:
on hold until we merge history management.

PR #450, opened 2023-05-03
Add conversation API, introduce `Exchange` history

calyptobai:
In this PR, we add an sqlite database for storing conversation history. We also add a few endpoints for managing this history, namely listing, deleting, and retrieving conversations.

The `ResponseState` struct has been simplified and refactored into the idea of an `Exchange`. Exchanges are stored directly in the corresponding `Conversation` for retrieval at a later date.

## Using `sqlx` with the database

To create new queries or modify existing ones, we can use `sqlx`. You'll need to `cargo install sqlx-cli` first, then these commands can be run from within `server/bleep`:

- `sqlx database reset --database-url=sqlite://bleep.db`: This will set up the database from scratch, deleting an old database if it exists
- `cargo sqlx prepare --database-url=sqlite://bleep.db`: This will update the `sqlx-data.json` file, which is nominally used to build the app without requiring a database instance.

### DIFF
--- a/server/bleep/src/webserver.rs
+++ b/server/bleep/src/webserver.rs
@@ -65,7 +65,14 @@ pub async fn start(app: Application) -> anyhow::Result<()> {
         .route("/search", get(semantic::complex_search))
         .route("/answer", get(answer::handle))
         .route("/answer/conversations", get(answer::conversations::list))
-        .route("/answer/conversations", delete(answer::conversations::delete));
+        .route(
+            "/answer/conversations/:thread_id",
+            get(answer::conversations::thread),
+        )
+        .route(
+            "/answer/conversations",
+            delete(answer::conversations::delete),
+        );
 
     if app.env.allow(Feature::AnyPathScan) {
         api = api.route("/repos/scan", get(repos::scan_local));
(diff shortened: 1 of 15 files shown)


server/bleep/src/webserver.rs
```diff
@@ -65,7 +66,16 @@ pub async fn start(app: Application) -> anyhow::Result<()> {
...
+        .route("/answer", get(answer::handle))
+        .route("/answer/conversations", get(answer::conversations::list))
+        .route(
+            "/answer/conversations/:thread_id",
+            get(answer::conversations::thread),
+        )
+        .route(
+            "/answer/conversations",
+            delete(answer::conversations::delete),
+        );
```
nerdypepper, line 78: the two routes for `/answer/conversations` can be merged as `.route("/answer/conversations", get(..).delete(..))`.

nerdypepper (approved):
works great, we should merge this in. 

couple of questions (just want to discuss these, no changes required here):

- we introduce a new `data_dir`, is there a reason we need a separation between the `index_dir` and the `data_dir`?
-  a conversation id is unique by means of `(user_id, thread_id)`, can we make that the primary key instead?

PR #456, opened 2023-05-03
Provision GH App credentials at the start of the service

rsdy:


ggordonhall:
This change does not resolve the issue

PR #459, opened 2023-05-04
improve error handling around `answer` step

nerdypepper:
an invalid response from the LLM results in the pipeline being interrupted with:

```
data:{
    "Err": "failed to parse `answer` response, expected array of arrays"
}
```


calyptobai (approved):
:+1:

PR #461, opened 2023-05-05
attach thread-id to response

nerdypepper:
also converts `thread_id` into a uuid everywhere. uuid::Uuid is copy and easy to throw around.

### DIFF
--- a/server/bleep/Cargo.toml
+++ b/server/bleep/Cargo.toml
@@ -105,7 +105,7 @@ tokenizers = "0.13.2"
 tokio-stream = "0.1.12"
 ort = { git = "https://github.com/bloopai/ort", branch = "merge-upstream" }
 ndarray = "0.15"
-uuid = { version = "1.2.2", features = ["v4", "fast-rng"] }
+uuid = { version = "1.2.2", features = ["v4", "fast-rng", "serde"] }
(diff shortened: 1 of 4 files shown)


overdrivenpotato:
Rather than attaching a `thread_id` to every `Exchange` instance, can we instead create a `stream::once(...)` [here](https://github.com/BloopAI/bloop/pull/461/files#diff-545e7a23fe176f3befb1953e53c8f0b70d7f62e99358da11ffd270fc7db8a02cL127-L133) containing the thread ID, and then chain the remaining exchange messages onto the stream afterwards? Much like we do with `[DONE]` at the end. This will avoid dealing with the thread ID on each exchange object.

calyptobai:
Rather than attaching a `thread_id` to every `Exchange` instance, can we instead create a `stream::once(...)` [here](https://github.com/BloopAI/bloop/pull/461/files#diff-545e7a23fe176f3befb1953e53c8f0b70d7f62e99358da11ffd270fc7db8a02cL127-L133) containing the thread ID, and then chain the remaining exchange messages onto the stream afterwards? Much like we do with `[DONE]` at the end. This will avoid dealing with the thread ID on each exchange object.

calyptobai (approved):
:+1:

PR #464, opened 2023-05-05
Improve repo sync state logic

rsdy:
This should also increase interactivity on the backend, and schedule re-indexing quicker on errors.


ggordonhall (approved):
lgtm

PR #472, opened 2023-05-09
load previous exchange's summary into next query

nerdypepper:
this patchset loads up a summarized version of the previous exchange in this conversation as an 'assistant' prompt. if the previous exchange is organized as:

    ["cite", 0, "This is the model/ directory"]
    ["cite", 1, "This is the gpt-2/ directory"]
    ["con", "The model files are present in the model/ directory"]

then the generated summary is produced by resolving path aliases and stripping json content:

    "This is the model/ directory" in "model/"
    "This is the gpt-2/ directory" in "model/gpt-2/"
    "The model files are present in the model/ directory"

the goal of this patchset is to allow answering queries that refer to previous citations or conclusions, in this example, the query "what is the parent directory of this directory?" is understood by the llm to be refering to the `model/` directory.

### DIFF
--- a/server/bleep/src/webserver/answer.rs
+++ b/server/bleep/src/webserver/answer.rs
@@ -23,7 +23,7 @@ use futures::{
 use reqwest::StatusCode;
 use secrecy::ExposeSecret;
 use tokio::sync::mpsc::Sender;
-use tracing::trace;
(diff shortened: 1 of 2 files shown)


calyptobai (approved):
:+1:

PR #475, opened 2023-05-10
Implement new context structure

calyptobai:
Needs a rebase before this can be merged.


server/bleep/src/webserver/answer.rs
```diff
@@ -503,16 +527,48 @@ impl Conversation {
...
+                s += &format!("### path alias: {} ###\n{snippet}\n\n", chunk.alias);
+            }
+
+            s
+        };
+
+        let query = self
+            .exchanges
+            .last()
+            .context("the exchange list was empty")?
```
nerdypepper, line 566: for a new conversation, the exchange list will always be empty, no?

calyptobai: Good catch, this is fixed now

PR #488, opened 2023-05-12
allow citing directories

nerdypepper:


server/bleep/src/webserver/answer/prompts.rs
```diff
@@ -148,6 +148,13 @@ pub fn final_explanation_prompt(context: &str, query: &str, query_history: &str)
             note: "This object can occur multiple times",
             example: None,
         },
+        Rule {
+            title: "Cite a single directory from the codebase",
+            description: "When you wish to cite every file in a directory, use this to directly cite the directory instead.",
+            schema: "[\"dir\",STRING: PATH]",
```
ggordonhall, line 154:
```suggestion
            schema: "[\"dir\",PATH:STRING]",
```

server/bleep/src/webserver/answer/prompts.rs
```diff
@@ -148,6 +148,13 @@ pub fn final_explanation_prompt(context: &str, query: &str, query_history: &str)
...
+            note: "This object can occur multiple times",
+            example: None,
```
ggordonhall, line 156: Could give an example here

ggordonhall (approved):

PR #491, opened 2023-05-12
Expand minimum line range in proc, merge overlapping ranges

calyptobai:


server/bleep/src/webserver/answer.rs
```diff
@@ -534,18 +541,31 @@ impl Conversation {
...
-                        "end": end,
-                        "relevant_code": lines
-                            .get(r.start.saturating_sub(1)..end.saturating_sub(1))?
+                .map(|mut r| {
+                    r.end = r.end.min(r.start + 20);
+                    r
+                })
+                .fold(Vec::<Range>::new(), |mut exps, next| {
+                    if let Some(prev) = exps.last_mut() {
+                        if prev.end >= next.start {
```
ggordonhall, line 552: We should merge ranges if the next range is within 10 lines of the next range. Something like:

```suggestion
                        if prev.end + 10 >= next.start {
```

ggordonhall (approved):

PR #499, opened 2023-05-15
introduce new `import` node

nerdypepper:
`import`s are both definitions (the introduction of a name to a syntax tree) and references (they refere to a name outside this syntax tree).

this should help improve cross-file navigation using the following heuristics:

- the definition of a name can exclude imports in cross-file navigation by default
- if no definition exists for a name across the entire repo, this name is an externally definied name, we can mark its import node in the same syntax-tree (or file) as its definition node


ggordonhall (approved):
On the whole this works well. There are still some cases that we need to think about, but can take these on in future PRs.

PR #513, opened 2023-05-17
relax schema requirements on final explanation

nerdypepper:
typically required the final explanation to be an array of arrays, such as:

    [
      ["cite", ...],
      ["dir", ...],
      ["con", ...],
    ]

this is now relaxed to permit a single item:

    ["con", ...]

### DIFF
--- a/server/bleep/src/webserver/answer.rs
+++ b/server/bleep/src/webserver/answer.rs
@@ -748,12 +748,11 @@ impl Conversation {
                 .ok_or(anyhow!("failed to parse `answer` response, expected array"))?;
 
             let array_of_arrays = json_array
+                .clone()
                 .into_iter()
                 .map(as_array)
                 .collect::<Option<Vec<Vec<_>>>>()
-                .ok_or(anyhow!(
-                    "failed to parse `answer` response, expected array of arrays"
-                ))?;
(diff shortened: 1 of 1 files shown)


server/bleep/src/webserver/answer.rs
```diff
@@ -748,12 +748,11 @@ impl Conversation {
...
 
             let array_of_arrays = json_array
+                .clone()
                 .into_iter()
                 .map(as_array)
                 .collect::<Option<Vec<Vec<_>>>>()
-                .ok_or(anyhow!(
-                    "failed to parse `answer` response, expected array of arrays"
-                ))?;
+                .unwrap_or_else(|| vec![json_array]);
```
ggordonhall, line 755: What error will we get if the model generates a string, so we end up with a `Vec<String>` here?

ggordonhall (approved):

server/bleep/src/webserver/answer.rs
nerdypepper: this scenario is permitted (it may recover as more tokens are streamed), however there can be other deviations from schema that are unrecoverable, added logging for those situations.

ggordonhall (approved):

PR #526, opened 2023-05-19
Store chunks retrieved in proc

calyptobai:
Now, we store chunks retrieved by `proc` in `code_chunks`.

### DIFF
--- a/server/bleep/src/webserver/answer.rs
+++ b/server/bleep/src/webserver/answer.rs
@@ -513,9 +513,8 @@ impl Conversation {
                     .await
                     .with_context(|| format!("failed to read path: {path}"))?
                     .content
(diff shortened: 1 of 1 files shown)


ggordonhall (changes requested):
This change will add enumerated chunks of code to `self.code_chunks`. These will then be re-enumerated, leaving us with chunks like

```
### path alias: 1 ###
40 40       analytics?.track('Repos Selected', {
41 41         localRepos,
42 42         githubRepos,
43 43         publicGithubRepos,
44 44         where,
45 45       });
46 46     },
47 47     [analytics],
```

We should add the un-enumerated lines to `self.code_chunks` here.

ggordonhall (approved):

PR #573, opened 2023-05-26
conditonally retry requests to LLM gateway

nerdypepper:
- do not retry on 400s
- do retry on 429s

to test this change set:

apply this patch to answer-api, and make a request to bleep, confirm that the LLM gateway attempts to retry 5 times:

    --- a/src/main.rs
    +++ b/src/main.rs
    @@ -212,6 +212,7 @@ async fn query(
         State(state): State<AppState>,
         Json(mut req): Json<answer_api::Request>,
     ) -> Result<Sse<impl Stream<Item = Result<Event, serde_json::Error>>>, (StatusCode, Error)> {
    +    return Err((StatusCode::TOO_MANY_REQUESTS, Error::BadOpenAiRequest));
         req.extra_stop_sequences.extend(
             [
                 "=========".to_owned(),

revert the previous patch and apply this patch to answer-api, make a request to bleep now and confirm that the LLM gateway does not attempt to retry at all:

    --- a/src/main.rs
    +++ b/src/main.rs
    @@ -212,6 +212,7 @@ async fn query(
         State(state): State<AppState>,
         Json(mut req): Json<answer_api::Request>,
     ) -> Result<Sse<impl Stream<Item = Result<Event, serde_json::Error>>>, (StatusCode, Error)> {
    +    return Err((StatusCode::BAD_REQUEST, Error::BadOpenAiRequest));
         req.extra_stop_sequences.extend(
             [
                 "=========".to_owned(),

### DIFF
--- a/server/bleep/src/webserver/answer/llm_gateway.rs
+++ b/server/bleep/src/webserver/answer/llm_gateway.rs
@@ -3,6 +3,7 @@
 use std::time::Duration;
 
 use anyhow::bail;
+use axum::http::StatusCode;
 use futures::{Stream, StreamExt};
 use reqwest_eventsource::EventSource;
 use tracing::warn;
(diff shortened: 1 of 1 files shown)


calyptobai (commented):
Can we make a `ChatError` here instead of reducing to a boolean in the oneshot function? Something like this would allow us to preserve the original error, and it would be more friendly for marking the function public if we decide to ever do so:

```rust
enum ChatError {
    BadRequest,
    TooManyRequests,
    Other(anyhow::Error),
}
```

calyptobai (approved):

PR #575, opened 2023-05-26
Track LLM gateway errors

calyptobai:
Due to our integration of `sentry_tracing`, `error!` logs are captured as sentry errors. We propagate additional information like message history using breadcrumbs generated by `debug!`.

### DIFF
--- a/server/bleep/src/webserver/answer/llm_gateway.rs
+++ b/server/bleep/src/webserver/answer/llm_gateway.rs
@@ -6,7 +6,7 @@ use anyhow::{anyhow, bail};
 use axum::http::StatusCode;
 use futures::{Stream, StreamExt};
 use reqwest_eventsource::EventSource;
-use tracing::warn;
+use tracing::{debug, warn};
 
 pub mod api {
     #[derive(serde::Serialize, serde::Deserialize, Debug, Clone, PartialEq)]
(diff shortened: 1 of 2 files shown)


ggordonhall (approved):

ggordonhall:
Import `error` then lgtm

server/bleep/src/webserver/answer/llm_gateway.rs
```diff
@@ -143,16 +143,18 @@ impl Client {
...
             match self.chat_oneshot(messages).await {
                 Err(ChatError::TooManyRequests) => {
-                    warn!("LLM request failed, retrying ...");
+                    warn!("too many LLM requests, retrying in {delay:?}...");
                     tokio::time::sleep(delay).await;
                     delay = Duration::from_millis((delay.as_millis() as f32 * SCALE_FACTOR) as u64);
                 }
                 Err(ChatError::BadRequest) => {
-                    warn!("LLM request failed, request not eligible for retry");
+                    debug!("LLM message list: {messages:?}");
```
rsdy, line 151: It would be more idiomatic to use `error!(?messages, "LLM request failed")` instead of an additional `debug!`.

server/bleep/src/webserver/answer/llm_gateway.rs
```diff
@@ -143,16 +143,18 @@ impl Client {
         for _ in 0..self.max_retries {
             match self.chat_oneshot(messages).await {
                 Err(ChatError::TooManyRequests) => {
-                    warn!("LLM request failed, retrying ...");
+                    warn!("too many LLM requests, retrying in {delay:?}...");
```
rsdy, line 146: `warn!(?delay, "LLM request failed, retrying with delay...")` would be nicer

rsdy (commented):
I think the style could be improved. It may also be helpful to [`instrument`](https://docs.rs/tracing/latest/tracing/attr.instrument.html) the whole function to get scoped errors for debugging.

server/bleep/src/webserver/answer/llm_gateway.rs
calyptobai: Wouldn't this cause the message list to log as a debug string to the console every time this error occurs? This means we will get large error strings even when `BLOOP_LOG=bleep=debug` is *not* set, or similar.

rsdy: This is true, it would be helpful to add a comment about this, as it's a bit unintuitive.

PR #583, opened 2023-05-30
update rectify_str to work with multibyte chars

nerdypepper:
rectify_str assumes that the width of a character is exactly 1 byte. this changeset introduces a new `advance` function that advances a string slice by 1 character, while being agnostic of character width.

this changeset also updates the `consume` function to be multibyte wary.


calyptobai (approved):
:+1:

PR #585, opened 2023-05-30
fix fuzzy-path search on multibyte strings

nerdypepper:
iterate over char indices instead of byte indices while constructing the regex filter in fuzzy-path search.


ggordonhall (changes requested):
Can you add some unit tests for fuzzy path search?

PR #593, opened 2023-05-31
Update action selection prompt

ggordonhall:
- Update action selection prompt to encourage bloop to choose a tool and not to directly respond to the question
- Persist discovered paths in the system prompt. Before bloop could hallucinate path aliases because they had been `[HIDDEN]`

### DIFF
--- a/server/bleep/src/webserver/answer/prompts.rs
+++ b/server/bleep/src/webserver/answer/prompts.rs
@@ -1,102 +1,86 @@
-//! Various prompts passed to the LLM.
-
-pub const INITIAL_PROMPT: &str = "Hi there, what can I help you with?";
+use std::fmt;
 
 pub const CONTINUE: &str = "Is there anything else I can help with?";
 
-pub fn system() -> String {
-    let today = chrono::offset::Local::now();
-    let tools = serde_json::json!([
-      {
-        "title": "Semantic code search",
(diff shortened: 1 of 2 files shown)


server/bleep/src/webserver/answer/prompts.rs
```diff
@@ -1,102 +1,86 @@
...
-              }
-          ],
-        },
-      }
-    ]);
+      Tool {
+        name: "processFiles", 
+        description: "Read one or more files and extract the line ranges which are relevant to the search terms.", 
+        schema: "{\"name\": \"proc\", \"args\": [SEARCH_TERMS // str, ARRAY_OF_PATH_ALIASES // int[]}", 
+        examples: "[[\"proc\", \"find all the functional react components\", [2, 5]], [\"proc\", \"where are error types\", [0]], [\"proc\", \"gitoxide initialisation\", [2, 5, 8]]]"
```
calyptobai, line 41: Do we want the `Tool::name` to match the `name` given in `Tool::schema`? In this case here, we see `processFiles`, but in the schema and examples, the tool is named `proc`.

calyptobai (approved):

server/bleep/src/webserver/answer/prompts.rs
ggordonhall: Not necessarily. We want the `name` field to be descriptive, and the `name` in `Tool::schema` to be short.

PR #595, opened 2023-06-01
Index unbranched HEADs

rsdy:


ggordonhall (approved):
Makes sense, and seems to work well.

PR #601, opened 2023-06-02
add `is_directory` flag to file index

nerdypepper:
introduces a boolean flag to the file index that indicates if the document is a directory or a file.

### DIFF
--- a/server/bleep/src/indexes/file.rs
+++ b/server/bleep/src/indexes/file.rs
@@ -545,7 +545,7 @@ impl File {
             RepoDirEntry::Dir(dir) => {
                 trace!("writing dir document");
                 let doc = dir.build_document(
-                    &self,
+                    self,
                     repo_name,
                     relative_path.as_path(),
                     repo_disk_path,
@@ -562,7 +562,7 @@ impl File {
                 let buf_size = file.buffer.len();
                 let doc = file
                     .build_document(
-                        &self,
+                        self,
                         repo_name,
                         relative_path.as_path(),
                         repo_disk_path,
@@ -612,6 +612,7 @@ impl File {
 }
 
 impl RepoDir {
+    #[allow(clippy::too_many_arguments)]
     fn build_document(
         self,
         schema: &File,
@@ -651,6 +652,7 @@ impl RepoDir {
 }
 
 impl RepoFile {
+    #[allow(clippy::too_many_arguments)]
     fn build_document(
         mut self,
         schema: &File,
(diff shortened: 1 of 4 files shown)


nerdypepper:
of note: this changes the index schema and will trigger a re-index (for both tantivy and qdrant).

server/bleep/src/repo/iterator/fs.rs
```diff
@@ -39,37 +39,32 @@ impl FileSource for FileWalker {
...
                             warn!(%err, ?entry_disk_path, "read failed; skipping");
                             return None;
                         }
                         Ok(buffer) => buffer,
-                    }
+                    };
+                    Some(RepoDirEntry::File(RepoFile {
+                        buffer,
+                        path: entry_disk_path.to_string_lossy().to_string(),
+                        branches: vec!["head".into()],
```
rsdy, line 58: can we make this `"head"` into a static in `super`?

`static HEAD: &str = "HEAD";`

server/bleep/src/indexes/file.rs
```diff
@@ -484,44 +490,40 @@ impl Indexer<File> {
...
-        } else {
-            relative_path.to_string_lossy().to_string()
-        };
-        trace!("processing file");
-
         let content_hash = {
             let mut hash = blake3::Hasher::new();
             hash.update(crate::state::SCHEMA_VERSION.as_bytes());
-            hash.update(file.buffer.as_bytes());
+            hash.update(dir_entry.buffer().unwrap_or_default().as_bytes());
```
rsdy, line 522: The issue here is that the content of all directories will be the same, therefore the hash of all directories will be the same. It may be a good idea to include a filename in this hash as well to make sure they're all unique.

server/bleep/src/indexes/file.rs
```diff
@@ -537,109 +539,135 @@ impl File {
...
-            match scope_graph {
-                // we have a graph, use that
-                Ok(graph) => SymbolLocations::TreeSitter(graph),
-                // no graph, it's empty
-                Err(err) => {
-                    warn!(?err, %lang_str, "failed to build scope graph");
-                    SymbolLocations::Empty
-                }
+        match dir_entry {
+            RepoDirEntry::Dir(dir) => {
```
rsdy, line 545: Can you extract these into separate methods on `RepoDirEntry` for each enum variant?
It would possibly help our testing story to verify the logic dependencies here outside of this massive function body.

Maybe rust-analyzer's extract function is a good start separating these out.

rsdy (commented):
It looks like a good start, but I think this PR is a good place to introduce some test stubs to see how a minimum cut would work, and take a few more steps in this direction.

server/bleep/src/indexes/file.rs
nerdypepper: the file cache datastructure is already a map between a path and its content hash. which means that there is no benefit in adding the path to the content-hash: varying the path will create a new entry in the file cache altogether and not update the existing hash.

rsdy (approved):

PR #611, opened 2023-06-06
Separate schema from logic, include semantic schema in DB versioning

rsdy:


server/bleep/build.rs
```diff
@@ -21,9 +21,10 @@ fn main() {
 fn set_index_version() {
     use std::fs::{read_dir, read_to_string};
 
-    let model_directories = &["src/indexes", "src/intelligence/scope_resolution"];
+    let model_directories = &["src/intelligence/scope_resolution"];
     let model_files = &[
-        "src/indexes.rs",
+        "src/semantic/schema.rs",
+        "src/indexes/schema.rs",
```
nerdypepper, line 27: changes to, say `File::new` (currently in `src/indexes/file.rs`), such as modifying the index options of a field, could also require a re-index.

server/bleep/build.rs
```diff
@@ -21,9 +21,10 @@ fn main() {
 fn set_index_version() {
     use std::fs::{read_dir, read_to_string};
 
-    let model_directories = &["src/indexes", "src/intelligence/scope_resolution"];
+    let model_directories = &["src/intelligence/scope_resolution"];
     let model_files = &[
-        "src/indexes.rs",
+        "src/semantic/schema.rs",
```
nerdypepper, line 26: i believe the same holds for the semantic module as well, changes to the chunker (currently in `src/semantic/chunk.rs`) should trigger a re-index.

nerdypepper (changes requested):
the rest of the PR lgtm.

calyptobai (approved):
LGTM :+1:

PR #639, opened 2023-06-14
Multilevel caches

rsdy:
This PR improves on our index cache situation by making it symbolic, use an SQL backend, and granular enough to track itra-file differences, speeding up the indexing process for small changesets or metadata changes.

List of changes:

 * Add new branch-aware file-level cache
 * Recognize and update existing embeddings in qdrant
 * Add branch selector to UI with ability to sync new branches
 * Status reporting for incremental & branch indexing
 * Make all the querying API surface aware of branches
 * Ensure consistent query behaviour for filters specified in an LLM conversation.


server/bleep/migrations/20230613143506_file-cache.sql
```diff
@@ -0,0 +1,10 @@
+-- Add migration script here
+CREATE TABLE file_cache (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    repo_ref TEXT NOT NULL,
+    file_name TEXT NOT NULL,
+    hash TEXT NOT NULL,
```
ggordonhall, line 6: Hash of what?

server/bleep/src/db.rs
```diff
@@ -9,8 +9,7 @@ use crate::Configuration;
 pub type SqlDb = Arc<SqlitePool>;
 
 pub async fn init(config: &Configuration) -> Result<SqlitePool> {
-    fs::create_dir_all(&config.data_dir)?;
-    let data_dir = config.data_dir.to_string_lossy();
+    let data_dir = config.index_dir.to_string_lossy();
```
ggordonhall, line 12: Why remove `create_dir_all`?

server/bleep/src/indexes.rs
```diff
@@ -66,18 +68,14 @@ impl<'a> GlobalWriteHandle<'a> {
...
 
-        tokio::task::block_in_place(|| {
-            self.handles
-                .par_iter()
-                .map(|handle| {
-                    handle.index(&sync_handle.reporef, repo, &metadata, sync_handle.pipes())
-                })
-                .collect::<Result<Vec<_>, _>>()
-        })?;
+        futures::future::join_all(self.handles.iter().map(|handle| {
```
ggordonhall, line 73: Why the change here?

server/bleep/src/indexes/file.rs
```diff
@@ -637,6 +627,36 @@ impl RepoFile {
...
+        }
+        Entry::Occupied(mut val) => {
+            _ = val.insert((content_hash, branch_list.to_owned()).into());
+        }
+        Entry::Vacant(val) => {
+            _ = val.insert_entry((content_hash, branch_list.to_owned()).into());
+        }
+    }
+
+    trace!("cache miss");
```
ggordonhall, line 656: When would this happen?

server/bleep/src/semantic.rs
```diff
@@ -65,29 +67,38 @@ macro_rules! val_parse_str(($hash:ident, $val:expr) => {
...
             ..
         } = orig;
 
-        let mut converted = payload
-            .into_iter()
-            .map(|(key, value)| (key, kind_to_value(value.kind)))
-            .collect::<HashMap<String, serde_json::Value>>();
+        let Some(PointId { point_id_options: Some(PointIdOptions::Uuid(id)) }) = id
+	else {
+	    // unless the db was corrupted/written by someone else,
```
ggordonhall, line 82: By `db` are we referring to `Qdrant` here?

server/bleep/src/cache.rs
```diff
@@ -0,0 +1,260 @@
...
+
+impl<'a> ChunkCache<'a> {
+    pub async fn for_file(
+        qdrant: &'a QdrantClient,
+        repo_ref: &'a str,
+        repo_name: &'a str,
+        relative_path: &'a str,
+    ) -> anyhow::Result<ChunkCache<'a>> {
+        let response = qdrant
+            .search_points(&SearchPoints {
```
ggordonhall, line 152: Can we search Qdrant by `PointId` rather than by vector here? Where the `id` is a hash of identifying fields.

rsdy: No, because we don't know the identifying fields for chunks that are not currently being indexed. It's a null-vector that we're using for search, because this is the only way we can filter for attributes.

server/bleep/src/indexes/file.rs
rsdy: When the item isn't in the cache.

server/bleep/src/db.rs
rsdy: Since `data_dir` is redundant, we can rely on `index_dir` existing at the time this is created.

server/bleep/src/cache.rs
nerdypepper: if i understand, this search operation on qdrant produces all the points for a given file? can we use [`QdrantClient::scroll`](https://docs.rs/qdrant-client/latest/qdrant_client/client/struct.QdrantClient.html#method.scroll) instead?

server/bleep/src/cache.rs
```diff
@@ -0,0 +1,260 @@
...
+        &self,
+        data: &'a str,
+        embedder: impl FnOnce(&'a str) -> anyhow::Result<Embedding>,
+        mut payload: Payload,
+    ) -> anyhow::Result<()> {
+        let id = {
+            let mut bytes = [0; 16];
+            bytes.copy_from_slice(&blake3::hash(data.as_ref()).as_bytes()[16..32]);
+            Uuid::from_bytes(bytes).to_string()
+        };
```
nerdypepper, line 202: does this reduce a 32-byte hash to a 16-byte uuid? how resistant is this to collisions?

nerdypepper (changes requested):

server/bleep/src/indexes.rs
rsdy: It turned into async.

server/bleep/src/semantic.rs
rsdy: Yes.

server/bleep/src/cache.rs
```diff
@@ -0,0 +1,260 @@
...
+            .execute(&mut tx)
+            .await?;
+        }
+
+        tx.commit().await?;
+
+        Ok(())
+    }
+
+    pub(crate) async fn delete_for_repo(&self, reporef: &RepoRef) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
rsdy: The naming may be misleading, `delete_all` works on a `Transaction` and it's an implementation detail, `delete_for_repo` is public and creates a new transaction.

server/bleep/src/cache.rs
rsdy: Since `blake3` is a cryptographic hash, it should be pretty resistant to collisions. Reducing the 256bit output to 128 bit would still require more than 2^64 points in the db for a 50% chance of collision (according to the birthday paradox), so I don't think this is something we need to worry about too much.

server/bleep/src/indexes/file.rs
```diff
@@ -399,28 +390,21 @@ impl File {
...
-                _ = val.insert(content_hash.into());
-            }
-            Entry::Vacant(val) => {
-                _ = val.insert_entry(content_hash.into());
-            }
+        if matches!(dir_entry, RepoDirEntry::File(_))
+            && is_cache_fresh(cache, &unique_hash, &entry_pathbuf, &dir_entry)
+        {
+            info!("fresh; skipping");
+            return Ok(());
```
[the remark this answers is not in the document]
rsdy: Good spot, thanks.

nerdypepper (approved):
lgtm. will like to test this a bit.

server/bleep/src/cache.rs
```diff
@@ -0,0 +1,280 @@
...
+    }
+}
+
+impl<T> From<T> for FreshValue<T> {
+    fn from(value: T) -> Self {
+        Self { fresh: true, value }
+    }
+}
+
+pub(crate) type RepoCacheSnapshot = Arc<scc::HashMap<String, FreshValue<()>>>;
```
ggordonhall, line 52: Can you add comments explaining what these different levels of caching do?
- `RepoCacheSnapshot`
- `FileCache`
- `ChunkCache`

server/bleep/src/indexes/file.rs
```diff
@@ -44,26 +44,27 @@ struct Workload<'a> {
...
         &self,
         reporef: &RepoRef,
         repo: &Repository,
         repo_metadata: &RepoMetadata,
         writer: &IndexWriter,
         pipes: &SyncPipes,
     ) -> Result<()> {
-        let file_cache = repo.open_file_cache(&self.config.index_dir);
+        let file_cache = FileCache::new(&self.sql);
+        let repo_cache = file_cache.for_repo(reporef).await?;
```
ggordonhall, line 62: This is confusingly named as it's still a file cache. Can you rename to `file_cache_by_repo`

server/bleep/src/indexes/file.rs
```diff
@@ -194,17 +171,33 @@ impl Indexer<File> {
         // hits is a mapping between a document address and the number of trigrams in it that
         // matched the query
         let repo_ref_term = Term::from_field_text(self.source.repo_ref, &repo_ref.to_string());
+        let branch_term = branch
+            .map(|b| {
+                trigrams(b)
```
ggordonhall, line 176: Would this match two branches which have the same set of trigrams in a different order?

server/bleep/src/indexes/file.rs
```diff
@@ -285,20 +279,26 @@ impl Indexer<File> {
         // XXX: can we use the bloop query language here instead?
         let query_parser = QueryParser::for_index(
             file_index,
-            vec![self.source.repo_disk_path, self.source.relative_path],
+            vec![self.source.repo_ref, self.source.relative_path],
```
ggordonhall, line 282: Why this change?

server/bleep/src/indexes/file.rs
```diff
@@ -338,17 +338,34 @@ impl Indexer<File> {
...
-        let repo_query = Box::new(TermQuery::new(
+        query.push(Box::new(TermQuery::new(
             Term::from_field_text(self.source.repo_ref, &repo_ref.to_string()),
             IndexRecordOption::Basic,
-        ));
+        )) as Box<dyn Query>);
+
+        let branch_term = branch
+            .map(|b| {
+                trigrams(b)
```
ggordonhall, line 356: Again, would this match two branches with the same set of trigrams?

server/bleep/src/indexes/file.rs
```diff
@@ -408,42 +423,40 @@ impl File {
         };
         let entry_pathbuf = repo_disk_path.join(&relative_path);
 
-        let content_hash = {
+        let semantic_hash = {
```
ggordonhall, line 426: Do we need this `semantic_hash` if we select by `repo_name` and `relative_path` in Qdrant?

server/bleep/src/indexes/file.rs
```diff
@@ -523,13 +538,13 @@ impl RepoDir {
...
                 schema.raw_relative_path => relative_path_str.as_bytes(),
                 schema.repo_disk_path => repo_disk_path.to_string_lossy().as_ref(),
-                schema.entry_disk_path => entry_pathbuf.to_string_lossy().as_ref(),
                 schema.relative_path => relative_path_str,
                 schema.repo_ref => repo_ref,
                 schema.repo_name => repo_name,
                 schema.last_commit_unix_seconds => last_commit,
                 schema.branches => branches,
                 schema.is_directory => true,
+                schema.unique_hash => tantivy_cache_key,
```
ggordonhall, line 547: Can you give this a more descriptive name?

server/bleep/src/semantic.rs
```diff
@@ -395,19 +430,21 @@ impl Semantic {
         Ok(deduplicate_snippets(results, vector, limit))
     }
 
+    #[allow(clippy::too_many_arguments)]
     #[tracing::instrument(skip(self, repo_ref, relative_path, buffer))]
     pub async fn insert_points_for_buffer(
         &self,
         repo_name: &str,
         repo_ref: &str,
+        unique_hash: &str,
```
ggordonhall, line 439: Not clear what this is a hash of

server/bleep/src/semantic.rs
```diff
@@ -420,73 +457,66 @@ impl Semantic {
...
-                        .into_qdrant(),
-                    }),
-                    Err(err) => {
-                        warn!(?err, %chunk_prefix, "embedding failed");
-                        None
-                    }
-                },
-            )
-            .collect::<Vec<_>>();
+        let embedder = |c: &str| {
```
ggordonhall, line 460: Do we need this?

server/bleep/src/semantic.rs
```diff
@@ -420,73 +457,66 @@ impl Semantic {
...
+            info!("generating embedding");
+            self.embed(c)
+        };
+        chunks.par_iter().for_each(|chunk| {
+            let data = format!(
+                "{repo_name}\t{relative_path}\n{unique_hash}\n{}",
```
ggordonhall, line 466: Why is `unique_hash` here? This could negatively impact retrieval.

server/bleep/src/webserver/answer/context.rs
```diff
@@ -0,0 +1,172 @@
...
+        debug!(?query, %self.thread_id, "executing semantic query");
+        self.app
+            .semantic
+            .as_ref()
+            .unwrap()
+            .search(&query, limit, offset, retrieve_more)
+            .await
+    }
+
+    pub(super) async fn file_search(&self, path: &str) -> Result<ContentDocument> {
```
ggordonhall, line 127:
```suggestion
    pub(super) async fn get_file_contents(&self, path: &str) -> Result<ContentDocument> {
```

ggordonhall (changes requested):

server/bleep/src/background.rs
```diff
@@ -223,7 +286,7 @@ impl BoundSyncQueue {
         self,
         reporef: RepoRef,
     ) -> anyhow::Result<SyncStatus> {
-        let handle = SyncHandle::new(self.0.clone(), reporef, self.1.progress.clone());
+        let handle = SyncHandle::new(self.0.clone(), reporef, self.1.progress.clone(), None);
```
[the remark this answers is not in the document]
rsdy: It's not a single-branch alternative, it's just re-indexing with the existing filters (ie, not changing anything other than Git pulling the repo). We track the difference of branch filters in a `SyncHandle` to help report accurate status to the frontend through the various endpoints that are exposed.

server/bleep/src/cache.rs
```diff
@@ -0,0 +1,280 @@
...
+    qdrant: &'a QdrantClient,
+    cache: scc::HashMap<String, FreshValue<Payload>>,
+    update: RwLock<Vec<(PointsSelector, qdrant_client::client::Payload)>>,
+    new: RwLock<Vec<PointStruct>>,
+}
+
+impl<'a> ChunkCache<'a> {
+    pub async fn for_file(
+        qdrant: &'a QdrantClient,
+        content_hash: &'a str,
```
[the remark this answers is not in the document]
rsdy: `repo_ref` and `relative_path` are not sufficient to track unique files in a git repository when we're aware of branches. There are 2 kinds of deterministic unique ids we use: one for tantivy, which is file-level, and one in qdrant which is chunk-level. How those IDs are composed/created is not particularly interesting or relevant at this layer, only that they're a unique reference to a file based in a content-addressed storage space.

server/bleep/src/cache.rs
```diff
@@ -0,0 +1,280 @@
...
+    ) -> anyhow::Result<ChunkCache<'a>> {
+        let response = qdrant
```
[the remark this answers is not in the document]
rsdy: We may be able to further limit the fields in the payload we retrieve to just the relevant fields. Upon another glance at the docs, their `set_payload` and `overwrite_payload` functions are separate, so we don't need to send the full payload back, and can selectively patch the different fields.

server/bleep/src/indexes/file.rs
rsdy: Moves our logic from disk-based to a purely symbolic space, consistent with removing IDs that carry semantic meaning.

server/bleep/src/indexes/file.rs
rsdy: As noted above, the combination of `repo_name` and `relative_path` is insufficient to uniquely identify resources when dealing with branches.

server/bleep/src/indexes/file.rs
rsdy: It's not intended to carry semantic meaning. It needs to fulfill 2 criteria:

 * db-wide unique
 * deterministic
 
Relying on or expressing any other incidental property these IDs may carry is not a good idea.

server/bleep/src/semantic.rs
rsdy: We lazily generate embeddings, and we need to pass *a* lambda or object ref either way. Now the main purpose is to allow logging from this callback with all the associated context in the logging span.

server/bleep/src/semantic.rs
rsdy: It's being used to pin the chunk to the containing file. I can't reason either way about retrieval performance (although I think this is in the *probably fine* territory), but I guess we could construct the reference hash differently, and achieve the same effect.

ggordonhall: We don't want to embed it. `repo_name` and `relative_path` are included in `data` to improve retrieval (by matching on these fields).

server/bleep/src/cache.rs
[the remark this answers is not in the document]
rsdy: The other nuance here is that this algo will not really impact cold indexing times, as the caches will be empty. The only overhead we rack up here is around calling qdrant, which we can't avoid without a dedicated fastpath if we somehow propagate that this is a cold index.

server/bleep/src/indexes/file.rs
rsdy: I've tried resolving this issue, but moving to a [whitespace tokenizer](https://github.com/BloopAI/bloop/commit/36722257b49570f0588135f5eabc4c6fca467879) triggers an infinite loop in tantivy that eats up all the memory. We've seen this bug in passing, but here it's reliable.

Will stick with this approach for now, and revisit this once we upgrade tantivy to 0.20+

nerdypepper: >Would this match two branches which have the same set of trigrams in a different order?

yes, fixable by turning branches to tokenize differently, or by applying a regex filter after the trigram matching work.

server/bleep/src/webserver/answer.rs
```diff
@@ -596,6 +616,9 @@ struct Agent {
     thread_id: uuid::Uuid,
     query_id: uuid::Uuid,
 
+    query: String,
+    branch: Option<String>,
```
calyptobai, line 620: Can we move `branch` to `Conversation`? If a single conversation can span multiple branches, switching between exchanges, then can we move it to `Exchange` instead?

server/bleep/src/webserver/answer.rs
```diff
@@ -596,6 +616,9 @@ struct Agent {
     thread_id: uuid::Uuid,
     query_id: uuid::Uuid,
 
+    query: String,
```
calyptobai, line 619: Do we actually need to record this at all? We have the `branch` already, can't we just add that in `semantic_query_params`?

Or, if we do want to include the other fields like language and path, can we store those directly like `branch`?

calyptobai (changes requested):
I've merged in `main`, and moved some data around. Mostly it is the same, several values from `AppContext` were moved to `Agent`. Have left a few comments about the data added.

server/bleep/src/webserver/answer.rs
rsdy: @ggordonhall ?

server/bleep/src/webserver/answer.rs
rsdy: The idea would have been to record _all_ filters we recognize in the ql, and pass them on. The raw string is stored because `SemanticQuery` is referential, and we can't clone it to owned. `branch` is a separate field so that we can pass it to tantivy queries.

It would be great if we could store an owned `SemanticQuery` or the sort so we don't need to break out the individual values (and re-construct them manually when anything changes), but I didn't have a good idea how to do that.

server/bleep/src/webserver/answer.rs
ggordonhall: A `Conversation` can span and switch between multiple branches, so we should move it to `Exchange`

server/bleep/src/webserver/answer.rs
calyptobai: In that case, can we store an owned `SemanticQuery` on the `Exchange`, in order to capture these other filters? As noted [above](https://github.com/BloopAI/bloop/pull/639#discussion_r1254099183), we know that `branch` is per-exchange. Rather than storing each filter separately on `Exchange`, this would entail storing an owned `SemanticQuery<'static>` (which encapsulates the `branch` set).

We can create an owned value of type `SemanticQuery<'static>` with a method like so:

```rust
impl SemanticQuery<'a> {
    pub fn into_owned(self) -> SemanticQuery<'static> {
        // Recursively create this method on all lifetime bearers, terminating by creating an
        // owned `Cow<'static, str>` in `Literal`.

        // ...
    }
}
```

And we can create `Cow<'static, str>` from `Cow<'a, str>` value like so:

```rust
fn example<'a>(cow: Cow<'a, str>) -> Cow<'static, str> {
    Cow::Owned(cow.into_owned())
}
```

server/bleep/src/ee.rs
```diff
@@ -0,0 +1 @@
+pub(crate) mod webserver;
```
calyptobai, line 1: Can we rename the `ee` module to something less opaque?

ggordonhall: `ee` is a bit of an industry standard. See: 
- https://github.com/windmill-labs/windmill/blob/main/backend/src/ee.rs
- https://github.com/strapi/strapi/tree/main/packages/core/admin/ee

server/bleep/src/webserver/answer.rs
ggordonhall: @calyptobai Have pushed this, can you check it out?

server/bleep/src/webserver/answer.rs
ggordonhall: @calyptobai If this is a quick change can you take this on now?

server/bleep/src/ee/webserver.rs
```diff
@@ -0,0 +1,36 @@
...
+
+#[derive(Deserialize)]
+pub(crate) struct RepositoryPatch {
+    branch_filter: BranchFilter,
+}
+
+/// Patch a repository with the given payload
+/// This will automatically trigger a sync
+//
+pub(crate) async fn patch_indexed(
```
ggordonhall, line 24: Could this be used for deleting as well as syncing branches? We should rename to something branch specific like `sync_branch` or `patch_branch` (if deletes might be supported).

ggordonhall (dismissed):

server/bleep/src/webserver/answer.rs
calyptobai: Resolved in [3d772c3](https://github.com/BloopAI/bloop/pull/639/commits/3d772c307205938ca18ea42c5ad0f34ed4ce7490)

server/bleep/src/webserver/answer.rs
calyptobai: I've slightly simplified this in [3d772c3](https://github.com/BloopAI/bloop/pull/639/commits/3d772c307205938ca18ea42c5ad0f34ed4ce7490), we now rely on the parsed `SemanticQuery` stored on `Exchange`.

server/bleep/src/background.rs
```diff
@@ -178,14 +225,40 @@ impl BoundSyncQueue {
...
+        self.1.queue.push(handle).await;
+        Ok(finished.recv_async().await?)
+    }
+
+    /// Index new branches specified in the `BranchFilter`.
+    ///
+    /// Unlike `sync_and_index`, this allows queueing multiple syncs
+    /// for the same repo multiple times to allow incrementally
+    /// syncing individual branches without restrictions.
+    pub(crate) async fn add_branches_for_repo(self, reporef: RepoRef, new_branches: BranchFilter) {
```
calyptobai, line 251: Can we remove this function? Cargo warns that it is unused.

calyptobai (changes requested):

ggordonhall (approved):

PR #670, opened 2023-06-21
Apply Qdrant quantisation and use on-disk storage

ggordonhall:
Applies two optimisations to Qdrant:

1) Int8 quantisation. This reduces the size of a Qdrant index with `bloop` indexed from 70Mb to 58Mb with no noticeable drop in search performance. See: https://qdrant.tech/documentation/guides/quantization/#setting-up-scalar-quantization

2) `on_disk` storage of payloads. Decreases RAM by slightly increasing response times: https://github.com/qdrant/qdrant/blob/8cb1a4e159c5de37662e55c29c870834e751f1c6/config/config.yaml#L15


ggordonhall:
Closing: quantisation does _not_ actually lead to any storage gains. Both the original and quantised vectors are stored on disk.

PR #671, opened 2023-06-21
Add new analytics events, introduce /vote endpoint

calyptobai:


anastasiya1155:
When is the identify call made? Preferably should be done after github login or on launch

anastasiya1155:
For upvote route we need query_id, but FE doesn't know it. It is not present in answer stream

server/bleep/src/webserver/answer.rs
```diff
@@ -44,6 +44,30 @@ use response::{Exchange, SearchResult, SearchStep, Update};
...
+#[derive(Clone, Debug, serde::Deserialize)]
+pub struct Vote {
+    pub positive: bool,
+    pub thread_id: uuid::Uuid,
+    pub query_id: uuid::Uuid,
+    pub repo_ref: Option<RepoRef>,
+}
+
+pub(super) async fn vote(
+    Query(params): Query<Vote>,
```
anastasiya1155, line 56: we shouldn't use query params for POST requests, we should parse request body instead

server/bleep/src/webserver/answer.rs
```diff
@@ -44,6 +44,30 @@ use response::{Exchange, SearchResult, SearchStep, Update};
 
 const TIMEOUT_SECS: u64 = 60;
 
+#[derive(Clone, Debug, serde::Deserialize)]
+pub struct Vote {
+    pub positive: bool,
+    pub thread_id: uuid::Uuid,
+    pub query_id: uuid::Uuid,
+    pub repo_ref: Option<RepoRef>,
+}
```
anastasiya1155, line 53: also, should accept and track in analytics comment (for downvotes)

anastasiya1155 (changes requested):

calyptobai:
@anastasiya1155 The identify call is now made on first login. Do we want it to also occur on a specific route, for users which are already logged in? It should now include the parameters you noted, though `user_id` *is* `user_name`.

Additionally:

> For upvote route we need query_id, but FE doesn't know it. It is not present in answer stream

The first message sent back from `/answer` has been changed from a thread ID string to an initialization object of the form:

```json
{
    "thread_id": "...",
    "query_id": "...",
}
```

> also, should accept (and track in analytics) the user comment left when downvoting the response

Now, `.../vote` replaces `positive: boolean` with a `feedback` parameter in the forms:

```json
{"type":"positive"}
```
```json
{"type":"negative","feedback":"..."}
```

anastasiya1155:
I still don't see the 'identify' call in the analytics panel. Are you sure it is triggered in Tauri? As far as I remember login flow on the cloud and in Tauri differs.

anastasiya1155:
I don't see any events for repo indexed/removed in the analytics console. fyi the FE uses `delete /repos/indexed` and `get /repos/sync` endpoints for managing repos

calyptobai:
> I still don't see the 'identify' call in the analytics panel. Are you sure it is triggered in Tauri? As far as I remember login flow on the cloud and in Tauri differs.

We should now track successful logins with `/remotes/github/login`.

> I don't see any events for repo indexed/removed in the analytics console. fyi the FE uses `delete /repos/indexed` and `get /repos/sync` endpoints for managing repos

Deletions are now tracked via `DELETE /repos/indexed`. Addition is tracked as part of `PUT /repos/indexed`.

calyptobai:
> We need to do some tweaks to identify call:
> 
> * device_id shouldn't have a trailing newline (previously was trimmed on the FE)
> * user_id should be the same as tracking_id from /config endpoint previously used by FE
> * github_username should be in traits

> I don't see events for repo adding/removing in the analytics console

These should now be addressed.

anastasiya1155 (approved):

PR #672, opened 2023-06-21
Gitpod integration

rsdy:


qdrant/config.yaml
```diff
@@ -0,0 +1,13 @@
+debug: false
```
ggordonhall, line 1: Can we re-use `QDRANT_CONFIG_TEMPLATE`?

ggordonhall (approved):

qdrant/config.yaml
rsdy: I think we can.

nerdypepper (approved):
lgtm! would want `BLOOP_LOG` set to `bleep=debug` in the environment.

PR #674, opened 2023-06-22
enable code-nav for all tokens

nerdypepper:


ggordonhall (approved):
This works well. LGTM 👍

PR #678, opened 2023-06-22
Fix ruby compile error

ggordonhall:
Add `hoverable_query` to the Ruby `TSLanguageConfig`. This likely does not cover all Ruby hoverables but is a quick fix. cc. @nerdypepper


calyptobai:
Merging this to fix builds on `main`

PR #679, opened 2023-06-22
Pin Rust toolchain, use Rustls

calyptobai:
Based on our `flake.nix` file, it seems we have the following runtime dependencies: `openssl.out rocksdb git zlib nsync onnxruntime14`. Judging from our own `Cargo.lock`, and examining qdrant's `Cargo.lock`, we know that:

- [`rocksdb` is built statically and not needed at runtime](https://github.com/rust-rocksdb/rust-rocksdb/blob/master/librocksdb-sys/Cargo.toml#L16)  (unless provided by environment variables, which we don't do in default builds)
- [`zlib` is not used as default features are deactivated](https://github.com/qdrant/qdrant/blob/master/lib/segment/Cargo.toml#L28)
- [We removed `libgit2` in favor of `gitoxide`](https://github.com/BloopAI/bloop/pull/432)
- [We pin `onnxruntime` versions to specific binaries provided by `ort`](https://github.com/BloopAI/bloop/pull/231)

We also add a Rust toolchain configuration to pin the toolchain at `1.70.0`, and ensure that `openssl-sys` is built statically using the `vendored` feature.

This removes all remaining non-platform-specific runtime dependencies except `nsync`, which does not seem to be linked into either `bloop` builds or `qdrant`'s builds.

### DIFF
--- a/server/bleep/Cargo.toml
+++ b/server/bleep/Cargo.toml
@@ -30,14 +30,14 @@ harness = false
 
 # core
 tantivy = { version = "0.19.2", features = ["mmap"] }
-tokio = { version = "1.28.2", features = ["macros", "process", "rt", "rt-multi-thread", "io-std", "io-util", "sync", "fs"] }
+tokio = { version = "1.29.0", features = ["macros", "process", "rt", "rt-multi-thread", "io-std", "io-util", "sync", "fs"] }
 futures = "0.3.28"
 rayon = "1.7.0"
-clap = { version = "4.3.4", features = ["derive"] }
+clap = { version = "4.3.9", features = ["derive"] }
 tracing = "0.1.37"
 tracing-subscriber = { version = "0.3.17", features = ["env-filter", "registry"] }
 color-eyre = "0.6.2"
-sqlx = { version = "0.6.3", features = ["sqlite", "migrate", "offline", "runtime-tokio-native-tls"] }
+sqlx = { version = "0.6.3", features = ["sqlite", "migrate", "offline", "runtime-tokio-rustls", "chrono"] }
 
 # for debugging
 console-subscriber = { version = "0.1.9", optional = true }
@@ -48,8 +48,8 @@ anyhow = "1.0.71"
 thiserror = "1.0.40"
 
 # query parsing
-pest = "2.6.0"
-pest_derive = "2.6.0"
+pest = "2.7.0"
+pest_derive = "2.7.0"
 
 # code-nav
 tree-sitter = "0.20.10"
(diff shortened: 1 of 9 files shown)


server/bleep/Cargo.toml
```diff
@@ -110,6 +110,9 @@ jsonwebtoken = { version = "8.3.0", features = ["use_pem"] }
 tiktoken-rs = "0.4.2"
 semver = { version = "1", features = ["serde"] }
 
+# vendored deps
+openssl-sys = { version = "0.9.88", features = ["vendored"] }
```
rsdy, line 114: We don't currently even transitively depend on `openssl`, so I don't see a reason why we would bring this dependency in.

I also think `vendored` openssl likely won't work for us, as it doesn't use the system cert store, so we'll need additional work to make HTTPS through it functional.

rsdy (changes requested):

server/bleep/Cargo.toml
calyptobai: > We don't currently even transitively depend on `openssl`, so I don't see a reason why we would bring this dependency in.

What do you mean?

```
openssl-sys v0.9.88
├── native-tls v0.2.11
│   ├── hyper-tls v0.5.0
│   │   └── reqwest v0.11.18
│   │       ├── bleep v0.4.5
...
```

> I also think `vendored` openssl likely won't work for us, as it doesn't use the system cert store, so we'll need additional work to make HTTPS through it functional.

Do you have an example of this failing? I've tested on linux and Windows, and it doesn't seem to have any issues. We can use `openssl-probe` if there are broken examples.

rsdy: Oh, in this case a better fix would be to just use `rustls` for `rustls-tls-native-roots` feature on `reqwest` and avoid openssl altogether.

rsdy: It seems like there are a few things going on here.
 1. OpenSSL is occasionally pulled in through `hyper-tls`, depending on host platform
 2. `bleep` pulled in a few changes recently that subtly add a dependency here, but these are easy to fix and switch to `rustls` based on `cargo tree` output.
 3. `tauri` currently only supports vendored and unvendored OpenSSL, and our app config uses the vendored version, so the library (I have to assume) will be correctly initialized by Tauri. However, their `dev` branch already supports rustls, so I have to assume sooner or later they'll release it.
 
 I think the meaningful way forward here is to make sure `bleep` builds without `hyper-tls`, and deal with Tauri when they release their new feature configuration.

calyptobai: As discussed offline, we can tackle this in another PR, given it seems certificates load correctly on all platforms

rsdy (approved):

PR #681, opened 2023-06-23
Only process the first 'max_tokens' of a file

ggordonhall:
- Only `proc` the first 16k tokens of each file
- Tokens after that are pruned


rsdy (approved):

nerdypepper (approved):
lgtm

PR #688, opened 2023-06-23
Add article format answer type

calyptobai:
We add a new answer format named "article", which is similar to the "conclusion" tool previously generated as part of `final_explanation_prompt`.


ggordonhall (approved):
Looks good to merge. We'll have to revisit this logic and clean it up.

PR #690, opened 2023-06-27
refactor code-nav

nerdypepper:
moves core-logic from the `webserver` to `intelligence`.


ggordonhall:
What are the FE changes on this branch?

nerdypepper:
@ggordonhall the response format changed a little, previously we were sending defs and refs separately, in order to mimic the structure required in the UI (two tabs: "Definitions" and "References"). now that the popup merges the two listings, the format on the BE follows in suit: a single listing of files, with a list of symbols in each file.

ggordonhall (approved):

PR #693, opened 2023-06-27
Add frequency penalty for proc queries

ggordonhall:
Occasionally the proc query ignores its instructions and generates hundreds of line ranges. This PR addresses this by setting a low `frequency_penalty` for proc calls, discouraging the model from generating many ranges. Calling proc with this penalty has been evaluated on a test set which shows that it does not degrade proc performance.

Also updates the action selection prompt, instructing the model to proc a maximum of 10 files at once.


server/bleep/src/webserver/answer/prompts.rs
```diff
@@ -31,7 +31,7 @@ pub fn functions() -> serde_json::Value {
             },
             {
                 "name": "proc",
-                "description": "Read one or more files and extract the line ranges which are relevant to the search terms.",
+                "description": "Read one or more files and extract the line ranges which are relevant to the search terms. Do not proc more than 10 files at a time.",
```
calyptobai, line 34: Why do we add this in the prompt rather than when consuming the results?

calyptobai (approved):

server/bleep/src/webserver/answer/prompts.rs
ggordonhall: The hypothesis is that it encourages the model to be selective about the files that it processes. Want to avoid situations where the model chooses to process every file in the path history. Have seen an example where it chose to process 54 files.

nerdypepper (approved):

nerdypepper (approved):

PR #694, opened 2023-06-27
Use native message history when generating LLM request

calyptobai:
Rather than embedding the query history as a string, we can create native `Message`s, with user and assistant roles respectively. This means that the prompt only changes depending on context and not on the message chain.

### DIFF
--- a/server/bleep/src/webserver/answer.rs
+++ b/server/bleep/src/webserver/answer.rs
@@ -923,7 +923,7 @@ impl Conversation {
         let system_message = prompts::final_explanation_prompt(&context, query);
         let messages = Some(llm_gateway::api::Message::system(&system_message))
             .into_iter()
-            .chain(query_history)
+            .chain(query_history.iter().cloned())
             .collect::<Vec<_>>();
 
         let mut stream = ctx.llm_gateway.chat(&messages, None).await?.boxed();
--- a/server/bleep/src/webserver/answer.rs
+++ b/server/bleep/src/webserver/answer.rs
@@ -294,7 +294,7 @@ impl Conversation {
         }
     }
 
-    fn query_history(&self) -> Vec<llm_gateway::api::Message> {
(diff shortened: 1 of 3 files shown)


server/bleep/src/webserver/answer.rs
```diff
@@ -915,15 +924,12 @@ impl Conversation {
...
-            .last_exchange()
-            .query()
-            .context("exchange did not have a user query")?;
-
-        let prompt = prompts::final_explanation_prompt(&context, query, &query_history);
-
-        let messages = [llm_gateway::api::Message::system(&prompt)];
+        let query_history = self.query_history().collect::<Vec<_>>();
+        let system_message = prompts::final_explanation_prompt(&context);
+        let messages = Some(llm_gateway::api::Message::system(&system_message))
```
ggordonhall, line 929: Is there a concern that we'll run out of prompt space if we put _all_ the query, answer pairs in the prompt? We could grab the last `n` (5?), or populate until we run out of prompt space.

calyptobai: In this case, we're using the conclusions only, which tend to be much shorter. Message contents should be identical compared to `main`.

ggordonhall: Would we be putting whole explanations in? These could be much longer than our current conclusions. We want to support cases where the agent explains how a function works, and the user responds `can you explain that in more detail`

calyptobai: To be clear, this PR does not add whole explanations to the history, for the token concerns you noted. We can do that, but this specifically is a refactoring change designed to help #688. We get better delegation between user and assistant messages.

Did you want to include whole explanations in this PR?

ggordonhall: Understood. Nah, let's add explanations in #688

ggordonhall (approved):

PR #699, opened 2023-06-29
fix bugs in hoverables queries

nerdypepper:
- scoped identifiers in rust result in `module::type` being hoverable, whereas we require `module` and `type` to be individually hoverable.
- catch clause variables in typescript were hoverable but did not produce results


ggordonhall (approved):
lgtm

PR #704, opened 2023-06-29
Catch non-existent paths in single file answer contexts

calyptobai:


server/bleep/src/indexes/file.rs
```diff
@@ -300,7 +300,7 @@ impl Indexer<File> {
         &self,
         query: Box<dyn tantivy::query::Query>,
         searcher: tantivy::Searcher,
-    ) -> Result<ContentDocument> {
+    ) -> Result<Option<ContentDocument>> {
```
nerdypepper, line 303: i think we could structure the return type a bit better in a future PR; `top_hit` (and its descendant functions `file_body` and `by_path`) can:
- return 0 docs if there are no results
- exactly 1 doc
- more than 1 doc if there are duplicates in the index

in this PR it is modeled as:
- `Ok(None)`: zero results
- `Ok(Some(_))`: one result
- `Err(_)`: many results

but this might be nicer if we use error enum variants for each scenario.

nerdypepper (approved):

PR #714, opened 2023-06-30
Windows builds 2

rsdy:
Would be helpful to get an acceptance test on all platforms before merging.


apps/desktop/src-tauri/src/backend.rs
```diff
@@ -37,11 +37,19 @@ where
...
+
+        bundled.dylib_dir = Some(if cfg!(target_os = "macos") {
             app.path_resolver()
                 .resolve_resource("dylibs")
-                .expect("missing `apps/desktop/src-tauri/dylibs`"),
-        );
+                .expect("missing `apps/desktop/src-tauri/dylibs`")
+                .parent()
+                .expect("invalid path")
+                .join("frameworks")
```
calyptobai, line 47: Won't this resolve e.g.

```
bloop.app/Contents/Resources/dylibs/
```

to this? :

```
bloop.app/Contents/Resources/frameworks
```

I would expect the `Frameworks` directory to be one level up, and with a capital `F`.

apps/desktop/src-tauri/build.rs
```diff
@@ -6,20 +6,56 @@ use std::{
...
+    )
+    .unwrap();
+}
+
+#[cfg(not(target_os = "macos"))]
+fn copy(profile_dir: &Path) {
+    let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap();
     let dylib_name = match target_os.as_str() {
         "linux" => Some("libonnxruntime.so"),
         "macos" => Some("libonnxruntime.dylib"),
```
calyptobai, line 61: Related to the `target_os` comment above, we can just use the check here instead. As it stands, this implementation of `copy` uses `CARGO_CFG_TARGET_OS`, but the `target_os = "macos"` version does not.

calyptobai (commented):
I've left a few comments regarding some of the internals. I can test this PR out on Windows after any changes to make sure that it continues to work

ggordonhall (dismissed):
The server does not start on the `aarch64-apple-darwin` build.

ggordonhall (approved):

PR #724, opened 2023-07-04
Introduce `Agent` struct, simplify `Exchange` data structure

calyptobai:
In this PR we simplify the `answer` module, and clean up the `Exchange`-related data types. Now, the `Agent` struct is used to drive conversations instead of `Conversation`, and it holds surrounding conversation context, such as thread IDs and LLM gateway clients.

Of note:

- `Exchange` has been reorganized and simplified
  - `Outcome` replaces the old `enum Results`
  - `finished` has been removed, and can now be inferred by the closing of a stream
  - `SearchResult` has been renamed to `FileAction`, and all associated `*Result` types were renamed to `*Action`
- `Conversation` methods related to agent tasks were split into a new `Agent` type
- Methods related to the LLM gateway were moved from `AppContext` to `llm_gateway::Client`
- The `AppContext` struct was merged into `Agent`
- The `CancellationTrack` struct was merged into `Agent`

### DIFF
--- a/server/bleep/src/webserver/answer/exchange.rs
+++ b/server/bleep/src/webserver/answer/exchange.rs
@@ -17,8 +17,8 @@ impl Exchange {
     pub fn apply_update(&mut self, update: Update) {
         match update {
             Update::Step(search_step) => self.search_steps.push(search_step),
-            Update::Filesystem(file_actions) => {
-                self.set_file_actions(file_actions);
+            Update::Filesystem(file_results) => {
+                self.set_file_results(file_results);
             }
             Update::Article(text) => {
                 let outcome = self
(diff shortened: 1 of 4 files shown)


server/bleep/src/webserver/answer/exchange.rs
```diff
@@ -183,28 +154,28 @@ pub enum SearchStep {
...
+    Conclude(String),
 }
 
 #[derive(serde::Serialize, serde::Deserialize, Debug, Clone)]
-pub enum SearchResult {
-    Cite(CiteResult),
-    Directory(DirectoryResult),
-    Modify(ModifyResult),
-    Conclude(ConcludeResult),
+pub enum FileAction {
```
ggordonhall, line 163: Not sure about the renaming here. Feel like this overlaps too much with the notion of Agent`Action`s. These don't really _do_ anything, so I prefer the `Result` terminology.

ggordonhall (approved):
Like this, nice and clean

PR #728, opened 2023-07-05
Generate XML code in article responses

calyptobai:
Code blocks are now generated in XML format, in an attempt to provide more accurate results.

We also prune code from XML blocks when generating an article summary.


server/bleep/src/lib.rs
```diff
@@ -4,10 +4,10 @@
     future_incompatible,
     nonstandard_style,
     rust_2018_idioms,
-    unused_crate_dependencies,
     unused_lifetimes,
     unused_qualifications
 )]
+#![warn(unused_crate_dependencies)]
```
rsdy, line 10: What is the crate that makes this necessary? Can we instead make it explicit down below for the config targets it affects?

calyptobai: This was originally `#[deny]`, but this is annoying during development. I was testing versions between parser crates, and at some point had a compilation error that I temporarily fixed by using `use crate_name as _;`. This is suboptimal for developer UX. Making this a warning means we still notice when there is an issue, but avoids the situation where WIP code doesn't compile due to ongoing changes.

PR #729, opened 2023-07-06
Generate article summary responses inline 

calyptobai:
We replace the previous fixed summaries with summaries generated inline
with the article prompt.

Some of the logic which makes use of the `comrak` crate is difficult to
follow. Comments have been added for clarity.

### DIFF
--- a/server/bleep/Cargo.toml
+++ b/server/bleep/Cargo.toml
@@ -123,6 +123,9 @@ sentry-tracing = "0.31.5"
 git-version = "0.3.5"
 gix = { version="0.47.0", features = ["blocking-http-transport-reqwest-rust-tls", "pack-cache-lru-static"] }
 thread-priority = "0.13.1"
+# We use the git version here, so that we can pull in recent changes that make footnotes work. The
+# latest crates.io version at the time of writing does not include necessary patches.
+comrak = { default-features = false, git = "https://github.com/kivikakk/comrak" }
 
 [target.'cfg(windows)'.dependencies]
 dunce = "1.0.4"
--- a/server/bleep/src/webserver/answer.rs
+++ b/server/bleep/src/webserver/answer.rs
@@ -1,4 +1,3 @@
-use rand::{seq::SliceRandom, SeedableRng};
 use secrecy::ExposeSecret;
 use std::{
     collections::{HashMap, HashSet, VecDeque},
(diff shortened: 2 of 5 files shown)


server/bleep/Cargo.toml
```diff
@@ -123,6 +123,9 @@ sentry-tracing = "0.31.5"
 git-version = "0.3.5"
 gix = { version="0.47.0", features = ["blocking-http-transport-reqwest-rust-tls", "pack-cache-lru-static"] }
 thread-priority = "0.13.1"
+# We use the git version here, so that we can pull in recent changes that make footnotes work. The
+# latest crates.io version at the time of writing does not include necessary patches.
+comrak = { default-features = false, git = "https://github.com/kivikakk/comrak" }
```
ggordonhall, line 128: Shouldn't we pin a commit?

server/bleep/src/webserver/answer.rs
```diff
@@ -1251,22 +1250,20 @@ impl Agent {
...
+                self.update(Update::Article(response.clone())).await?;
+            }
         }
 
-        let article_conclusion_messages = vec![
-            "I hope that was useful, can I help with anything else?",
-            "Is there anything else I can help you with?",
-            "Can I help you with anything else?",
-        ];
+        let summary = split_article_summary(&response)
```
ggordonhall, line 1262: Maybe we should backoff to a canned answer if the model doesn't generate a summary?

nerdypepper (approved):
lgtm!

ggordonhall (changes requested):
We should backoff to a canned response if the model does not generate a summary. This has happened a couple of times in testing.

server/bleep/src/webserver/answer.rs
calyptobai: Added as a fallback in 3cd6c943453329a32f078c5daa4390dd3d7860b8

calyptobai:
> Although fine for now, might https://github.com/wooorm/markdown-rs be a better choice of markdown parser?

This could work as well. Though looking at its documentation, it seems like it also closely follows `cmark-gfm`, much like `comrak`. I suspect that the footnote definition hack will still be necessary because of this. But we can try it before merging this if needed?

ggordonhall:
> > Although fine for now, might https://github.com/wooorm/markdown-rs be a better choice of markdown parser?
> 
> This could work as well. Though looking at its documentation, it seems like it also closely follows `cmark-gfm`, much like `comrak`. I suspect that the footnote definition hack will still be necessary because of this. But we can try it before merging this if needed?

No `comrak` is fine for now

PR #735, opened 2023-07-07
Re-introduce HyDE, work around batched calls

calyptobai:
We work around the issues present before, by avoiding `batch` searches in qdrant. It is not clear what the problem is currently, but those calls tend to throw HTTP2 stream errors when testing in cloud deployments. Given the volume of documents tends not to be very large, we can buffer a set of non-batched calls instead.


ggordonhall (approved):
Lgtm

PR #736, opened 2023-07-07
Expand code chunks to try and maximize context

calyptobai:
Depends on #780.

We now try to expand code chunks to fill 70% of the context, using a greedy iterative algorithm:

1. Calculate per-chunk expansion: we target a small line increment (100 lines as currently implemented), and divide this number to obtain per-chunk line increments.
2. We expand each chunk both up and down, according to this increment
3. We merge overlapping chunks
4. Repeat until we reach our 70% threshold

This iterative approach avoids problems with overlapping spans, and is fairly simple to follow.

### DIFF
--- a/server/bleep/src/webserver/answer.rs
+++ b/server/bleep/src/webserver/answer.rs
@@ -430,11 +430,10 @@ impl Agent {
         self.exchanges.last_mut().expect("exchange list was empty")
     }
 
-    fn code_chunks(&self) -> Vec<CodeChunk> {
+    fn code_chunks(&self) -> impl Iterator<Item = CodeChunk> + '_ {
         self.exchanges
             .iter()
             .flat_map(|e| e.code_chunks.iter().cloned())
-            .collect::<Vec<_>>()
     }
 
     fn paths(&self) -> Vec<String> {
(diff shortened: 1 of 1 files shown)


ggordonhall (dismissed):
Would like to run through this in more detail, but I think this should be restructured.

We should keep `canonicalise` as it was, and store these merged chunks in `self.code_chunks`. We should then expand chunks to fill the prompt on-the-fly in a separate method that's call in `answer_context`. `answer_context` takes the `path_aliases` of the files that GPT has deemed relevant to the query, so we only want to expand these. 

Can you add a brief description of the expansion algorithm?

calyptobai:
> Would like to run through this in more detail, but I think this should be restructured.
>
> We should keep `canonicalise` as it was, and store these merged chunks in `self.code_chunks`.

As discussed in a call, this has been rebased against #780, providing transient code chunks that prevent canonicalization from overwriting the old cached chunk list.

>  We should then expand chunks to fill the prompt on-the-fly in a separate method that's call in `answer_context`. `answer_context` takes the `path_aliases` of the files that GPT has deemed relevant to the query, so we only want to expand these.

This has now been added in 7f8dbec136eef62491effe47dba59a9ad9ae6dac.

> Can you add a brief description of the expansion algorithm?

The PR description has now been updated.

ggordonhall (approved):
LGTM 👍

PR #739, opened 2023-07-10
Sync-related consistency improvements

rsdy:
Fixes bugs where:
 * Upstream repository changes aren't consistently picked up by the indexer
 * Race condition where Bloop tries to update repositories before credentials are available
 * If more than the number of worker threads repositories are queued, they aren't immediately reflected on the UI, only when they become active in the sync queue.

### DIFF
--- a/server/bleep/src/background/sync.rs
+++ b/server/bleep/src/background/sync.rs
@@ -87,7 +87,7 @@ impl Drop for SyncHandle {
 }
 
 impl SyncHandle {
-    pub(crate) fn new(
+    pub(crate) async fn new(
         app: Application,
         reporef: RepoRef,
         status: super::ProgressStream,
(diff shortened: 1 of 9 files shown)


server/bleep/src/background/sync.rs
```diff
@@ -87,23 +87,46 @@ impl Drop for SyncHandle {
...
-            reporef,
             new_branch_filters,
             exited,
             exit_signal,
-        }
-        .into()
+        };
+
+        _ = sh
+            .create_new(|| {
```
ggordonhall, line 108: Can we just roll `create_new` into this method? Would make this logic easier to follow

server/bleep/src/periodic/remotes.rs
```diff
@@ -138,6 +138,10 @@ async fn update_credentials(app: &Application) {
 }
 
 pub(crate) async fn check_repo_updates(app: Application) {
+    while app.credentials.github().is_none() {
```
ggordonhall, line 141: Could we get stuck in this state?

server/bleep/src/repo/iterator/git.rs
```diff
@@ -58,7 +58,7 @@ impl GitWalker {
         let head_name = head
             .clone()
             .try_into_referent()
-            .map(|r| r.name().to_owned());
+            .map(|r| format!("origin/{}", human_readable_branch_name(&r)));
```
ggordonhall, line 61: Can you add a comment explaining why we do this?

server/bleep/src/periodic/remotes.rs
rsdy: Yes, if there's no authentication we'll never enter this loop.

ggordonhall (approved):

PR #749, opened 2023-07-12
Lowercase organisation name

ggordonhall:
We currently operate on cased GitHub organisation names. This is unnecessary as the GitHub API `org` parameter is case insensitive.


rsdy:
We probably also want to force this in the `RepoRef` parser code to stay consistent.

rsdy (approved):

PR #754, opened 2023-07-17
Expose `proc` function only if there are paths in context

ggordonhall:
Currently, certain queries (`Explain lines 1 - 16 in vimrc`) cause the agent to predict `proc` before it has discovered any paths. This PR only exposes `proc` as a function when there are paths in its context.


rsdy (approved):

stunningpixels:
Seems to fix

PR #757, opened 2023-07-17
Qdrant performance tuning

rsdy:
In Kube it seems like we need a bit more tolerance, as qdrant isn't as fast to consolidate writes as we'd like. Add random backoff to retries, wait for write confirmation during indexing, tune the qdrant WAL & keep more things in memory to get a faster resolution.

### DIFF
--- a/helm/bloop/values.yaml
+++ b/helm/bloop/values.yaml
@@ -191,7 +191,7 @@ qdrant:
       enabled: true
       p2p:
         port: 6335
-      concensus:
+      consensus:
         tick_period_ms: 100
     storage:
       snapshots_path: /qdrant/storage/snapshots
--- a/helm/bloop/values.yaml
+++ b/helm/bloop/values.yaml
@@ -188,24 +188,11 @@ qdrant:
   config:
     telemetry_disabled: true
     cluster:
-      enabled: false
+      enabled: true
       p2p:
         port: 6335
       consensus:
         tick_period_ms: 100
     storage:
+      async_scorer: true
       snapshots_path: /qdrant/storage/snapshots
-      on_disk_payload: true
-      wal:
-        wal_capacity_mb: 64
-        wal_segments_ahead: 8
-      hnsw_index:
-        max_indexing_threads: 4
-
-      optimizers:
-        max_optimization_threads: 4
-        max_segment_size_kb: 16384
-        memmap_threshold_kb: 4096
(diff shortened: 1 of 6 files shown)


helm/bloop/values.yaml
```diff
@@ -185,11 +186,19 @@ qdrant:
...
     cluster:
       enabled: true
       p2p:
         port: 6335
-      consensus:
+      concensus:
         tick_period_ms: 100
     storage:
       snapshots_path: /qdrant/storage/snapshots
+      on_disk_payload: false
```
ggordonhall, line 198: What's the effect of this on RAM? Not sure I understand this comment in Qdrant's config template: https://github.com/qdrant/qdrant/blob/7164ac4a5987d28f1c93f5712aef8e09e7d93555/config/config.yaml#L18

server/bleep/src/cache.rs
```diff
@@ -169,7 +169,7 @@ impl<'a> ChunkCache<'a> {
         let response = qdrant
             .scroll(&ScrollPoints {
                 collection_name: semantic::COLLECTION_NAME.to_string(),
-                limit: Some(1_000_000),
+                limit: Some(1_000),
```
ggordonhall, line 172: How does `limit` affect `Scroll` performance?

server/bleep/src/semantic.rs
```diff
@@ -467,10 +468,28 @@ impl Semantic {
         branches: &[String],
         is_cold_run: bool,
     ) {
-        let chunk_cache =
-            crate::cache::ChunkCache::for_file(&self.qdrant, tantivy_cache_key, is_cold_run)
-                .await
-                .expect("qdrant error");
+        let chunk_cache = 'cache: {
+            let rand = rand::distributions::Uniform::new(500, 1500);
```
ggordonhall, line 472: Why is the backoff random? Can you add a comment explaining what's going on here and why

helm/bloop/values.yaml
rsdy: Will probably increase RAM pressure significantly on large repos.

server/bleep/src/cache.rs
rsdy: It cuts off potential huge right tail here. In practice we don't really have more than 10s of chunks per file from my testing, so it's really just to keep things sensible.

PR #759, opened 2023-07-18
Fix local repo syncing

rsdy:
Caused by #739, this logic is now in one place.


ggordonhall (approved):
Seems to work now, lgtm

PR #763, opened 2023-07-20
Only trigger cleanups a day after restart

rsdy:
This is more of a temp fix for the bug where bloop unsyncs branch indexes, and it's not great. A better tracking of branch additions & query times is necessary, however, for improved logic.


ggordonhall (approved):
lgtm

PR #766, opened 2023-07-21
Use sqlite for chunk cache

rsdy:
The PR moves the chunk caching mechanism adjacent to qdrant in order to avoid querying it during intensive write cycles.
Before this patch we would need to wait for qdrant to become queryable periodically, stalling our own processing pipeline, which is problematic since the indexing operation on both sides is fairly expensive (and thus can take a lot of time). Therefore chunk caches now live in SQLite.

The chunks during a write operation are updated in parallel both in qdrant and SQLite.

If the sqlite db is deleted, it is possible that an inconsistency will leave stale objects in qdrant. Since chunk ids in qdrant are deterministic, an up-to-date cache will be built back up, although recalculating all embeddings will be necessary.

### DIFF
--- a/server/bleep/migrations/20230613143506_file-cache.sql
+++ b/server/bleep/migrations/20230613143506_file-cache.sql
@@ -1,6 +1,11 @@
 -- Add migration script here
 CREATE TABLE file_cache (
-    id INTEGER PRIMARY KEY AUTOINCREMENT,
-    repo_ref TEXT NOT NULL,
-    cache_hash TEXT NOT NULL
+    cache_hash TEXT PRIMARY KEY NOT NULL,
+    repo_ref TEXT NOT NULL
+);
+
+CREATE TABLE chunk_cache (
+    chunk_hash TEXT NOT NULL,
+    file_hash TEXT NOT NULL,
+    branches TEXT NOT NULL
 );
--- a/server/bleep/migrations/20230613143506_file-cache.sql
+++ b/server/bleep/migrations/20230613143506_file-cache.sql
@@ -7,5 +7,6 @@ CREATE TABLE file_cache (
 CREATE TABLE chunk_cache (
     chunk_hash TEXT NOT NULL,
     file_hash TEXT NOT NULL,
-    branches TEXT NOT NULL
+    branches TEXT NOT NULL,
+    repo_ref TEXT NOT NULL
 );
--- a/server/bleep/src/cache.rs
+++ b/server/bleep/src/cache.rs
@@ -5,7 +5,7 @@ use qdrant_client::{
     qdrant::{PointId, PointStruct},
 };
 use sqlx::Sqlite;
-use tracing::{debug, trace};
+use tracing::trace;
 use uuid::Uuid;
 
 use crate::{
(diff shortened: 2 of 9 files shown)


server/bleep/migrations/20230613143506_file-cache.sql
```diff
@@ -1,6 +1,12 @@
...
-    id INTEGER PRIMARY KEY AUTOINCREMENT,
-    repo_ref TEXT NOT NULL,
-    cache_hash TEXT NOT NULL
+    cache_hash TEXT PRIMARY KEY NOT NULL,
+    repo_ref TEXT NOT NULL
+);
+
+CREATE TABLE chunk_cache (
+    chunk_hash TEXT NOT NULL,
+    file_hash TEXT NOT NULL,
```
nerdypepper, line 9: is this a foreign key to the `file_cache` table?

rsdy: No, these are different. This is also why we need the `repo_ref`, so we can cross-reference which ones we need to keep when a repo is being deleted.

server/bleep/src/cache.rs
```diff
@@ -221,93 +197,170 @@ impl<'a> ChunkCache<'a> {
...
+                        .map(PointId::from)
+                        .collect::<Vec<_>>()
+                        .into(),
+                    None,
+                )
                 .await?;
         }
 
+        //
+        // Upsert points
```
ggordonhall, line 321: Why is this an `upsert`? Aren't these just new points?

server/bleep/src/cache.rs
```diff
@@ -221,93 +197,170 @@ impl<'a> ChunkCache<'a> {
...
+        //
         let new: Vec<_> = std::mem::take(self.new.write().unwrap().as_mut());
         let new_size = new.len();
+        let new_sql = std::mem::take(&mut *self.new_sql.write().unwrap());
```
ggordonhall, line 325: Can we rename this to `new_chunk_cache_entries` or something?

ggordonhall (changes requested):
I like this!

server/bleep/src/cache.rs
rsdy: They should be, but this operation has always been `upsert`, so I didn't change it. We don't really require strict semantics that ensure these be new, all we're saying is "we want these to exist like this".

server/bleep/src/cache.rs
```diff
@@ -118,18 +112,15 @@ impl<'a> FileCache<'a> {
...
         Ok(())
     }
 
-    async fn delete_tx(
-        &self,
-        tx: &mut sqlx::Transaction<'_, Sqlite>,
-        reporef: &RepoRef,
-    ) -> anyhow::Result<()> {
-        let repo_str = reporef.to_string();
+    async fn delete_tx(&self, tx: &mut sqlx::Transaction<'_, Sqlite>) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
rsdy: The entire `FileCache` instance operates on a repository level.

ggordonhall (approved):

PR #785, opened 2023-07-26
Add `/answer/explain` route for explaining file ranges

calyptobai:
This new route allows the client to "explain" a line range in a file, and requires less conversation turns due to pre-generated conversation history.


server/bleep/src/webserver/answer.rs
```diff
@@ -91,17 +93,70 @@ fn default_thread_id() -> uuid::Uuid {
...
+    let response = execute_agent(
+        params.clone(),
+        app.clone(),
+        user.clone(),
         query_id,
+        conversation_id,
+        exchanges,
+        action,
     )
     .await;
```
ggordonhall, line 162: By moving logic up and removing `_handle`, won't we now fail to track errors that are thrown by any logic outside of `execute_agent`? We want visibility over these errors.

server/bleep/src/webserver/answer.rs
```diff
@@ -332,6 +345,103 @@ pub(super) async fn _handle(
...
+#[derive(serde::Deserialize)]
+pub struct Explain {
+    pub relative_path: String,
+    pub line_start: usize,
+    pub line_end: usize,
+    pub branch: Option<String>,
+    pub repo_ref: RepoRef,
+    #[serde(default = "default_thread_id")]
+    pub thread_id: uuid::Uuid,
+}
```
ggordonhall, line 357: With this change we're handling the requests very differently across `/answer` and `/explain`. In `/answer` we're using the bloop query language to specify branch information, whereas here we're explicitly passing as a param.

We could simplify things by using the bloop query language for both types of request, and using the same `Params` struct that we used to. The client can build a query of the form: `Explain lines {} - {} in {} branch:<> path:<>` and pass as `q`. This saves us the need for the `virtual_req`.

server/bleep/src/webserver/answer.rs
```diff
@@ -332,6 +345,103 @@ pub(super) async fn _handle(
...
+            .branch
+            .insert(Literal::Plain(std::borrow::Cow::Owned(branch)));
+    }
+
+    let file_content = app
+        .indexes
+        .file
+        .by_path(&virtual_req.repo_ref, &params.relative_path, None)
+        .await
+        .context("failed to conduct path search")?
```
ggordonhall, line 403: These error messages would confuse me. I'd think we'd explicitly done a path search. Can you rephrase so more clearly conveys that we've tried and failed to fetch a file's content?

ggordonhall (changes requested):

server/bleep/src/webserver/answer.rs
calyptobai: This is now patched in ca68093, by *always* tracking errors automatically in `execute_agent`.

server/bleep/src/webserver/answer.rs
calyptobai: As discussed offline, we can't merge the two as `/answer/explain` needs to pre-fetch file data to synthesize a code chunk, and needs direct access to line numbers to construct a focused chunk.

ggordonhall (approved):

PR #786, opened 2023-07-27
add file metadata to open responses

nerdypepper:
- `size`: size of the file in bytes
- `loc`: number of lines in the file

these fields are present only on responses with `kind` set to 'file'.

### DIFF
--- a/server/bleep/src/query/execute.rs
+++ b/server/bleep/src/query/execute.rs
@@ -159,6 +159,8 @@ pub struct FileData {
     lang: Option<String>,
     contents: String,
     siblings: Vec<DirEntry>,
+    size: usize,
+    loc: usize,
 }
 
 #[derive(Serialize)]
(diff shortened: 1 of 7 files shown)


server/bleep/src/query/execute.rs
```diff
@@ -671,6 +673,8 @@ impl ExecuteQuery for OpenReader {
                         repo_ref: doc.repo_ref.to_owned(),
                         lang: doc.lang.clone(),
                         contents: doc.content.clone(),
+                        size: doc.content.len(),
+                        loc: doc.line_end_indices.len(),
```
anastasiya1155, line 677: can this number include only lines of code, i.e. not empty?

nerdypepper: added a new `sloc` field that measures non-blank lines.

anastasiya1155 (approved):

rsdy (approved):

PR #788, opened 2023-07-27
Local repositories without remotes were incorrectly indexed

rsdy:
This resulted in branch filters messing up queries and making it impossible to get results from purely local repos.

In this patch, the detection of whether or not it's a Git repository is changed from looking for a Git remote (which may not exist), to simply checking if it's a Git repository at all by signaling if we have detected a last commit timestamp.

Additionally, add explicit support local repositories in the Git file iterator so we only force remote branches when it's a remote repository.


nerdypepper (approved):
lgtm, fixes bug in indexing repos without remotes.

PR #789, opened 2023-07-27
Fix file browsing on Windows

calyptobai:
To enforce consistency, we now use forward slashes (`/`) throughout the app, on both unix and Windows platforms.


nerdypepper:
the `File` index contains an `is_directory` field, could we use that to classify documents instead of the trailing slash notation?

nerdypepper (approved):
lgtm, comments can be tackled in a later PR.

PR #799, opened 2023-08-01
Disable filesystem responses

ggordonhall:
Deprecate `filesystem` type answers. Now respond with `Article` type explanations for all queries. Remove legacy logic related to `filesystem` answering and rendering.


server/bleep/src/webserver/answer/exchange.rs
```diff
@@ -18,7 +18,7 @@ use crate::webserver::answer;
 pub struct Exchange {
     pub id: uuid::Uuid,
     pub query: SemanticQuery<'static>,
-    pub outcome: Option<Outcome>,
```
calyptobai, line 21: Can we rename `outcome` here to `answer`?

calyptobai (approved):

PR #802, opened 2023-08-02
log into files in addition to stderr

nerdypepper:
this changeset introduces persistent logging through the `tracing-appender` crate. logs are written into
`<index_dir>/logs/bloop.log.YYYY-MM-DD-HH` and rotated hourly. logs written to files are limited to the `bleep` crate.

### DIFF
--- a/server/bleep/src/lib.rs
+++ b/server/bleep/src/lib.rs
@@ -34,7 +34,10 @@ use once_cell::sync::OnceCell;
 use sentry_tracing::{EventFilter, SentryLayer};
 use std::{path::Path, sync::Arc};
 use tracing::{debug, error, info, warn, Level};
-use tracing_subscriber::EnvFilter;
+use tracing_subscriber::{
+    filter::{LevelFilter, Targets},
+    EnvFilter,
+};
 
 mod background;
 mod cache;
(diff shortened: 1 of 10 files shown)


server/bleep/src/lib.rs
```diff
@@ -338,25 +344,32 @@ impl FromRef<Application> for axum_extra::extract::cookie::Key {
...
+    let log_writer_layer = (!config.disable_log_write).then(|| {
+        let log_dir = config.index_dir.join("logs");
+        let file_appender = tracing_appender::rolling::hourly(log_dir, "bloop.log");
+        let (non_blocking, guard) = tracing_appender::non_blocking(file_appender);
+        _ = LOGGER_GUARD.set(guard);
+        fmt::layer()
+            .with_writer(non_blocking)
+            .with_ansi(false)
+            .with_filter(Targets::new().with_target("bleep", LevelFilter::DEBUG))
+            .with_filter(Targets::new().with_target("bleep::indexes::file", LevelFilter::WARN))
```
nerdypepper, line 359: `bleep::indexes::file` is the noisiest module of them all. we can limit this to `warn` and above, but have debug logs for the rest of the codebase (should give us insight into `webserver` just fine.)

ggordonhall: Can you recalculate the log size per hour with this change? After how many days should we cleanup logs?

nerdypepper:
to confirm that cloud instances do not log into files; get a shell into the dev-cluster and list contents of the index directory:
```
root@bloop-akshay-764465888-7zxbz:/# ls data/
index  lost+found  repos
```
the absence of the `logs` directory indicates that logs are not being stored on disk.

server/bleep/src/lib.rs
nerdypepper: ~130kB/hr; size shouldn't be a problem here, we can afford to keep logs for upto a month or more.

ggordonhall (approved):

PR #805, opened 2023-08-02
Store decoded responses directly on exchange, re-encode on the fly

calyptobai:
NOTE:

This PR swaps some notation. Where `encode` previously meant "encoding" XML-format articles into markdown for the front-end, we now call this `decode`, as we decode *from* XML-style articles to markdown. We also introduce the notion of `encode`, a reverse operation which takes markdown as *input* and generates an XML-style article.

---

We make a few big changes here:

1. LLM response parsing is moved to a new `transcoder` module
2. We rename `encode` to `decode`
3. We add a *new* `encode` function that takes markdown and generates
   XML-style LLM responses. This is used when recreating LLM history so
   that we can provide the model with correct syntax.
4. "Decoded" markdown is now stored directly on an exchange. We
   re-encode on the fly when generating LLM history.

This module also fixes some bugs related to parsing, as we *first* decode XML-style responses to markdown, before splitting an answer from its conclusion. Previously, this did not work well because we attempted to split an article body and summary *before* converting XML to markdown. Because our XML format is bespoke, it was not well understood
by markdown parsers. Splitting before conversion was done to ensure we store raw XML on the exchange, but given the new change to store *decoded markdown*, we can now avoid the issue entirely.

XML comments are now also stripped from LLM responses.


ggordonhall (approved):
Looks good!

PR #808, opened 2023-08-03
trim utterance history as necessary

nerdypepper:


server/bleep/src/webserver/answer.rs
```diff
@@ -949,14 +949,20 @@ impl Agent {
...
-        let history = self.utter_history().collect::<Vec<_>>();
-
-        let system_message = prompts::answer_article_prompt(&context);
-        let messages = Some(llm_gateway::api::Message::system(&system_message))
+        let system_prompt = prompts::answer_article_prompt(&context);
+        let system_message = llm_gateway::api::Message::system(&system_prompt);
+        let history = {
+            let h = self.utter_history().collect::<Vec<_>>();
+            let system_headroom =
+                tiktoken_rs::num_tokens_from_messages("gpt-4", &[(&system_message).into()])?;
```
ggordonhall, line 962:
```suggestion
                tiktoken_rs::num_tokens_from_messages(ANSWER_ARTICLE_MODEL, &[(&system_message).into()])?;
```

server/bleep/src/webserver/answer.rs
```diff
@@ -949,14 +949,20 @@ impl Agent {
...
+            trim_utter_history(h, ANSWER_HEADROOM + system_headroom)?
```
ggordonhall, line 963: We should pass `ANSWER_ARTICLE_MODEL` into `trim_utter_history` rather than hardcoding `gpt-4`.

server/bleep/src/webserver/answer.rs
```diff
@@ -44,6 +44,8 @@ use exchange::{Exchange, SearchStep, Update};
 use llm_gateway::api::FunctionCall;
 
 const TIMEOUT_SECS: u64 = 60;
+const ANSWER_MODEL: &str = "gpt-4";
+const ANSWER_MODEL_WITH_REV: &str = "gpt-4-0613";
```
calyptobai, line 48: Why do we have both of these constants, instead of just always using `gpt-4-0613`?

ggordonhall (approved):

server/bleep/src/webserver/answer.rs
nerdypepper: righto, looks like tiktoken looks for a prefix.

PR #809, opened 2023-08-03
Disable tauri file watching default

calyptobai:
Now, running `npm start` will boot an app that doesn't auto-recompile the Rust server when a source file is changed. To get back watching behaviour, it is now possible to run `npm run watch` instead.


ggordonhall (approved):
I'd approve this twice if I could

PR #811, opened 2023-08-03
Move agent-related code into new root `agent` module

calyptobai:
Webserver code that executes the agent remains in `webserver::answer`. Agent tool implementations and their supporting methods and functions were also moved into individual submodules.

Closes BLO-1341


ggordonhall (approved):
Can we also move `conversations.rs` up one into `webserver/`?

nerdypepper (approved):

PR #829, opened 2023-08-07
Reindex if SQL schema has changed

rsdy:


nerdypepper (changes requested):
i think we should add `migrations`, where the schema changes are reflected. `sqlx-data.json` only contains compiled sqlx queries (which is could also cause index format changes).

PR #830, opened 2023-08-08
Remove qdrant references

ggordonhall:
- Configure cloud instances to use central Qdrant instance
- Make `collectionName` configurable in `values.yaml`
- Create Qdrant field indices when creating a collection


server/bleep/src/semantic.rs
```diff
@@ -199,45 +184,73 @@ impl Semantic {
...
+                    .await?;
+                qdrant
+                    .create_field_index(
+                        &config.collection_name,
+                        "relative_path",
+                        FieldType::Text,
+                        None,
+                        None,
+                    )
+                    .await?;
```
rsdy, line 248: Note: these were outside of the `match` because we needed to migrate previously created collections to ones with indexes. Moving this inside the `match`  may subtly cause issues with upgrades if we rely on indexes being present in the future, if we e.g. choose to extend this code block.

Maybe a better way of dealing with any potential failures on this code path is splitting this section into a new function, loop over the indexed field names, and explicitly logging the results of these operations, but not cascading the failure.

ggordonhall: Could we just delete and re-create the collection on index migration?

rsdy: We could, but it's probably overkill. Adding or removing indexes doesn't necessarily mean the data inside is invalidated, just that we've found another/more efficient way to access it.

PR #855, opened 2023-08-14
Use Cognito web service to log in

rsdy:
This allows us to use a more principled permissioning system on our backends, but will require re-authentication of all users on upgrade.

UX & backend upgrades are required before merge.


server/bleep/src/webserver/github.rs
```diff
@@ -74,46 +74,82 @@ pub(super) async fn status(Extension(app): Extension<Application>) -> impl IntoR
 /// Connect to Github through OAuth Device Flow
```
ggordonhall, line 74: Update this description

server/bleep/src/webserver/github.rs
```diff
@@ -74,46 +74,82 @@ pub(super) async fn status(Extension(app): Extension<Application>) -> impl IntoR
...
+            ("client_id", &client_id),
+            ("state", &state),
+        ],
+    )
+    .unwrap()
+    .to_string();
+
+    json(GithubResponse::AuthenticationNeeded {
+        url,
+        code: "DEAD-BEEF".to_string(),
```
ggordonhall, line 117: Do we still need this `code`?

ggordonhall (dismissed):
Makes sense. There are some variables and comments that we'll need to update.

PR #862, opened 2023-08-16
Allow frontend to send backend logs on request.

rsdy:
The `get_last_log_file` Tauri call will return the last written log file to the frontend for error & crash reporting. We can expand this in the future.

### DIFF
--- a/apps/desktop/src-tauri/src/backend.rs
+++ b/apps/desktop/src-tauri/src/backend.rs
@@ -3,8 +3,10 @@ use std::sync::Arc;
 use bleep::{analytics, Application, Configuration, Environment};
 use once_cell::sync::OnceCell;
 use sentry::ClientInitGuard;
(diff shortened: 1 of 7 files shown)


apps/desktop/src-tauri/src/backend.rs
```diff
@@ -20,123 +22,192 @@ fn get_device_id() -> String {
 
 static SENTRY: OnceCell<ClientInitGuard> = OnceCell::new();
 
-pub(super) fn bleep<R>(app: &mut App<R>) -> plugin::Result<()>
+#[tauri::command]
+fn get_last_log_file(config: tauri::State<Configuration>) -> Option<String> {
```
anastasiya1155, line 26: to be able to call it, it has to be registered in apps/desktop/src-tauri/src/main.rs line 43

anastasiya1155 (approved):

PR #869, opened 2023-08-18
Make bloop more tolerant to network failures

rsdy:
There are 2 bugs fixed here:

1.  bloop wouldn't start without network connection
2.  we silence reporting Error conditions when we suspect it's stemming from the network (there's no exhaustive list of error states we know would be indicative of this, though)

1) requires the `credentials.json` migration, because now we persist the name of the user that's currently logged in (to validate if there's actually a logged in user). Originally the frontend blew up on start without network connection because `github/status` blew up because sentry blew up. So now all of this is channeled through `/config` which doens't blow up.

### DIFF
--- a/server/bleep/src/indexes/file.rs
+++ b/server/bleep/src/indexes/file.rs
@@ -532,7 +532,7 @@ impl RepoDir {
         relative_path: &Path,
         repo_disk_path: &Path,
         repo_ref: &str,
-        last_commit: i64,
+        last_commit: u64,
         tantivy_cache_key: String,
     ) -> tantivy::schema::Document {
         let relative_path_str = format!("{}/", relative_path.to_string_lossy());
@@ -577,7 +577,7 @@ impl RepoFile {
         tantivy_cache_key: String,
         entry_pathbuf: &Path,
         repo_ref: &str,
-        last_commit: i64,
+        last_commit: u64,
         repo_metadata: &RepoMetadata,
         file_cache: &FileCache,
     ) -> Option<tantivy::schema::Document> {
--- a/server/bleep/src/indexes/file.rs
+++ b/server/bleep/src/indexes/file.rs
@@ -532,7 +532,7 @@ impl RepoDir {
         relative_path: &Path,
         repo_disk_path: &Path,
         repo_ref: &str,
-        last_commit: u64,
+        last_commit: i64,
         tantivy_cache_key: String,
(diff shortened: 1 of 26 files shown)


ggordonhall (changes requested):
Dropping and restarting network connection while a repo is in state `Cloning` causes the repo to get stuck in the `Queued` state.

server/bleep/src/intelligence/language/python/mod.rs
```diff
@@ -159,10 +159,6 @@ mod tests {
                     imports: [
                         List {
                             context: "from typings import §List§",
```
ggordonhall, line 161: Why has this been removed?

server/bleep/src/indexes/file.rs
```diff
@@ -532,7 +532,7 @@ impl RepoDir {
         relative_path: &Path,
         repo_disk_path: &Path,
         repo_ref: &str,
-        last_commit: u64,
+        last_commit: i64,
```
ggordonhall, line 535: Why the sign change?

rsdy: It's a dependency requirement, gix changed fixed their apis to track the correct type for unix seconds. (previously it was different from the underlying POSIX APIs for some reason)

server/bleep/src/intelligence/language/python/mod.rs
rsdy: This is an expect test, this change is a result of upgrading the upstream deps.

nerdypepper: these tests shouldn't change on version upgrade. will take a look at what is going on here.

nerdypepper: can we roll-back the change to python version dependency? it contains a breaking change to the grammar.

server/bleep/src/state.rs
```diff
@@ -45,6 +40,10 @@ pub struct StateSource {
     #[clap(long)]
     #[serde(default)]
     cookie_key: Option<PathBuf>,
+
+    #[serde(skip)]
+    #[clap(skip)]
+    root_dir: PathBuf,
```
nerdypepper, line 46: what is root-dir? would be nice to have a doc-comment here.

nerdypepper (approved):

ggordonhall (approved):

PR #874, opened 2023-08-21
Move qdrant killing earlier

rsdy:
Move killing qdrant earlier in the event loop, as we sometimes seem to leave it around. Maybe moving it earlier in the shutdown process helps?


ggordonhall (commented):
Yes, I suspect that this might help. A Tauri dev recommended shutting down child processes on `ExitRequest`.

PR #875, opened 2023-08-21
Add output hash for git dependency esaxx-rs

Twey:
Without this hash, `nix build` will fail, as Cargo doesn't store hashes for git dependencies.


rsdy:
Yup, there has been some flux here. Thanks!

PR #878, opened 2023-08-21
Even better qdrant killing

rsdy:
Make the second attempt fail gracefully. There's not much point panicing here, as the app is supposed to be shutting down anyway.

It seems like `ExitRequested` is skipped on a Mac when using eg. CMD-Q to quit.

### DIFF
--- a/apps/desktop/src-tauri/src/qdrant.rs
+++ b/apps/desktop/src-tauri/src/qdrant.rs
@@ -5,6 +5,7 @@ use std::{
 };
 
 use tauri::{plugin::Plugin, Runtime};
+use tracing::warn;
 
 use super::relative_command_path;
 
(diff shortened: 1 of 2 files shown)


apps/desktop/src-tauri/src/qdrant.rs
```diff
@@ -49,11 +50,14 @@ where
     fn on_event(&mut self, _app: &tauri::AppHandle<R>, event: &tauri::RunEvent) {
         use tauri::RunEvent::{Exit, ExitRequested};
         if matches!(event, Exit | ExitRequested { .. }) {
-            self.child
-                .take()
-                .expect("qdrant not started")
-                .kill()
-                .expect("failed to kill qdrant")
+            let Some(mut child) = self.child.take() else {
+		warn!("qdrant has been killed");
```
ggordonhall, line 54:
```suggestion
		info!("qdrant has been killed");
```

ggordonhall (approved):

PR #881, opened 2023-08-22
add code-navigation utilites to code-studio 

nerdypepper:
- utility to extract the "value" of definitions:
  `ScopeGraph::value_of_definition` uses a heuristic to find a possible
  value for the definition. it is defined as the smallest surrounding
  scope or the largest adjacent scope. variable definitions or local
  definitions typically will not have an scope node "nearby", and their
  "value" can be assumed to span the containing line.

- add /related-files endpoint: uses scope-graphs to calculate a closure
  of files related to a target file. request and response schema are
  subject to change.

- add /token-value endpoint: exposes `ScopeGraph::value_of_definition`
  to the webserver. request and response schema are subject to change.

### DIFF
--- a/server/bleep/src/indexes/reader.rs
+++ b/server/bleep/src/indexes/reader.rs
@@ -28,6 +28,25 @@ pub struct ContentDocument {
     pub branches: Option<String>,
 }
 
+impl std::hash::Hash for ContentDocument {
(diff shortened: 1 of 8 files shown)


ggordonhall (approved):
LGTM 👍

PR #882, opened 2023-08-22
Allow batching of embeddings through server

rsdy:
For local deployments, we will keep the other options.

TODO:
 - [ ] helm charts


server/bleep/src/semantic/embedder.rs
```diff
@@ -0,0 +1,271 @@
...
+};
+use qdrant_client::qdrant::{PointId, PointStruct};
+use serde::{Deserialize, Serialize};
+use tokenizers::Tokenizer;
+use tracing::{error, info, trace};
+
+use super::Embedding;
+
+#[derive(Default)]
+pub struct EmbedLog {
```
ggordonhall, line 25: Find the naming here confusing, would prefer `EmbedQueue`.

server/bleep/src/semantic/embedder.rs
```diff
@@ -0,0 +1,271 @@
...
+        let sequence_embedding = &*output_tensor.view();
+        let pooled = sequence_embedding.mean_axis(Axis(1)).unwrap();
+        Ok(pooled.to_owned().as_slice().unwrap().to_vec())
+    }
+
+    fn tokenizer(&self) -> &Tokenizer {
+        &self.tokenizer
+    }
+
+    async fn batch_embed(&self, log: &EmbedLog, _flush: bool) -> anyhow::Result<Vec<PointStruct>> {
```
ggordonhall, line 148: Feels inconsistent that this returns `PointStruct`, should return `Vec<Embedding>` to mirror `self.embed`

server/bleep/src/semantic/embedder.rs
```diff
@@ -0,0 +1,271 @@
...
+    fn embed(&self, data: &str) -> anyhow::Result<Embedding> {
+        self.embedder.embed(data)
+    }
+
+    fn tokenizer(&self) -> &Tokenizer {
+        self.embedder.tokenizer()
+    }
+
+    async fn batch_embed(&self, log: &EmbedLog, flush: bool) -> anyhow::Result<Vec<PointStruct>> {
+        const MAX_BATCH_SIZE: usize = 128;
```
ggordonhall, line 203: This will change depending on server configuration. This should be a command line argument that is configured in Helm. Can use this as a default value.

server/bleep/src/semantic/embedder.rs
```diff
@@ -0,0 +1,271 @@
...
+impl Embedder for RemoteEmbedder {
+    fn embed(&self, data: &str) -> anyhow::Result<Embedding> {
+        self.embedder.embed(data)
+    }
+
+    fn tokenizer(&self) -> &Tokenizer {
+        self.embedder.tokenizer()
+    }
+
+    async fn batch_embed(&self, log: &EmbedLog, flush: bool) -> anyhow::Result<Vec<PointStruct>> {
```
ggordonhall, line 202: It's counter-intuitive that this method will only embed a batch if `log.len() >= MAX_BATCH_SIZE`. This should take a batch and embed it, and we should move the `EmbedLog` logic (and the construction of `PointStruct`) up into the `FileCache`.

server/bleep/src/semantic/embedder.rs
```diff
@@ -0,0 +1,271 @@
...
+    pub fn new(url: reqwest::Url, model_dir: &Path) -> anyhow::Result<Self> {
+        let url = url.join("encode")?;
+        Ok(Self {
+            url,
+            session: reqwest::Client::new(),
+            embedder: LocalEmbedder::new(model_dir)?,
+        })
+    }
+
+    async fn make_request(&self, request: ServerRequest<'_>) -> anyhow::Result<ServerResponse> {
```
ggordonhall, line 180: Can we just fold this method into `batch_embed`?

ggordonhall (changes requested):

server/bleep/src/cache.rs
```diff
@@ -61,11 +61,22 @@ pub(crate) type FileCacheSnapshot = Arc<scc::HashMap<String, FreshValue<()>>>;
 pub(crate) struct FileCache<'a> {
     db: &'a SqlDb,
     reporef: &'a RepoRef,
+    semantic: Option<&'a Semantic>,
```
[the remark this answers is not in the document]
rsdy: I didn't really want to go into this whole story here, as it would be a fairly invasive change. I'd rather handle this in a separate PR that's dedicated to this.

server/bleep/src/semantic/embedder.rs
rsdy: I've considered this. They can't have the same return type, since `EmbedLog` has all the details to create the `PointStruct`, and we're removing the items from the front of the queue, not just peeking into it (for consistency reasons).

It's significantly simpler separating and recombining the metadata with the data in one place, at least in this hierarchical setup. (which https://github.com/BloopAI/bloop/pull/882#discussion_r1310069663 would change)

server/bleep/src/semantic/embedder.rs
```diff
@@ -0,0 +1,271 @@
...
+                    output.extend(res.data.into_iter().zip(batch).map(|(result, src)| {
+                        PointStruct {
+                            id: Some(PointId::from(src.id)),
+                            vectors: Some(result.embedding.into()),
+                            payload: src.payload,
+                        }
+                    }))
+                }
+                Err(err) => {
+                    error!(
```
rmuller-ml, line 246: Can we call local embedder in case remote embedder raises an error. This will make it more reliable.

server/bleep/src/cache.rs
```diff
@@ -149,8 +163,102 @@ impl<'a> FileCache<'a> {
...
+                .upsert_points(&semantic.collection_name(), new_points, None)
+                .await
+            {
+                error!(?err, "failed to write new points into qdrant");
+            }
+        }
+        Ok(())
+    }
+
+    async fn embed_queued_points(
```
ggordonhall, line 191:
```suggestion
    async fn try_embed_queued_points(
```
Can we rename this, to make it clearer that this method will often do nothing?

server/bleep/src/cache.rs
```diff
@@ -160,18 +268,21 @@ impl<'a> FileCache<'a> {
...
 pub struct ChunkCache<'a> {
     sql: &'a SqlDb,
+    semantic: &'a Semantic,
     reporef: &'a RepoRef,
     file_cache_key: &'a str,
     cache: scc::HashMap<String, FreshValue<String>>,
     update: scc::HashMap<(Vec<String>, String), Vec<String>>,
-    new: RwLock<Vec<PointStruct>>,
     new_sql: RwLock<Vec<(String, String)>>,
+    embed_log: &'a EmbedQueue,
```
ggordonhall, line 277: Rename this `embed_queue`.

ggordonhall (approved):
Left a couple of naming nits, and we should make `RemoteEmbedder` `ee`, but apart from that LGTM

server/bleep/src/semantic/embedder.rs
ggordonhall: We'll tackle this in a separate PR

PR #883, opened 2023-08-23
Add fuzzy path search endpoint

ggordonhall:
This PR adds a path search endpoint `search/path` which can be called during the CS flow. It's fuzzy so will return paths that partially match the query too.

This change-set also simplifies `parse_nl` so that it always returns `SemanticQuery`. We weren't using `ParsedQuery::Grep` anywhere meaningful so this should simplify the logic.


server/bleep/src/query/execute.rs
```diff
@@ -44,6 +45,10 @@ pub struct ApiQuery {
     /// A query written in the bloop query language
     pub q: String,
 
+    /// Optional RepoRef to constrain the search. If not provided, search all repos
+    #[serde(default)]
+    pub repo_ref: Option<RepoRef>,
+
```
nerdypepper, line 51: don't we have `repo:xxx` for this?

server/bleep/src/webserver.rs
```diff
@@ -54,7 +54,8 @@ pub async fn start(app: Application) -> anyhow::Result<()> {
         .route("/hoverable", get(hoverable::handle))
         .route("/token-info", get(intelligence::handle))
         // misc
-        .route("/search", get(semantic::complex_search))
+        .route("/search/code", get(search::code_search))
+        .route("/search/path", get(search::path_search))
```
nerdypepper, line 58: can we clarify somewhere that these are semantic code search and fuzzy path search? if these endpoints are studio specific, maybe we should use a new route altogether, `studio/xxx`. @calyptobai what do you think?

server/bleep/src/webserver/search.rs
```diff
@@ -0,0 +1,85 @@
...
+            Err(Error::new(ErrorKind::UpstreamService, "error"))
+        }
+    }
+}
+
+pub(super) async fn path_search(
+    Query(args): Query<ApiQuery>,
+    Extension(indexes): Extension<Arc<Indexes>>,
+) -> impl IntoResponse {
+    match parser::parse_nl(&args.q.clone()) {
```
nerdypepper, line 40: we can avoid a `match` here with:

```
let q = parser::parse_nl(..).map_err(|e| {
    // Err arm here
})?;
// Ok arm here
```

nerdypepper (dismissed):

server/bleep/src/webserver.rs
ggordonhall: These endpoints are not strictly studio-specific. They can also be used to debug the `Agent` flow (they simulate the `code` and `path` tools).

PR #885, opened 2023-08-24
classify related-files into files-importing and files-imported

nerdypepper:
- files-imported: list of files imported in this file
- files-importing: list of files importing this file

these lists may have overlaps and may not be complete because they use just scope-graphs and not search-based navigation.


anastasiya1155:
It is working fine, although a bit slow and sometimes times out

anastasiya1155 (approved):

ggordonhall (approved):

PR #887, opened 2023-08-24
Migrate to on-disk qdrant collections

rsdy:


server/bleep/src/semantic/schema.rs
```diff
@@ -40,3 +48,18 @@ impl PartialEq for Payload {
         // created payload
     }
 }
+
+pub(super) fn create_collection(name: &str) -> CreateCollection {
```
nerdypepper, line 52: minor nit: can we have this share a signature similar to `create_indexes`: accept a reference to a qdrant client?

nerdypepper (approved):

nerdypepper (approved):

PR #896, opened 2023-08-29
Ensure consistency of semantic & tantivy & caches

rsdy:
Since the tracking of semantic & tantivy is done with different uniqueness constraints, the wrong key was used to delete stale chunks from qdrant, therefore nothing was deleted.

This PR unifies the handling of the 2 keys (`CacheKeys`), the tracking of changes, and defines one explicit synchronization point where everything needs to line up. All semantic db work now goes through the `ChunkCache`.

Overall this is an improvement to the mess we have in our file indexing logic, and creates better separation & dependency structure between the involved components. Some code moved around as a result. Maybe as next steps we should break out more files with the involved components to make navigation better.

Builds on #882.

### DIFF
--- a/server/bleep/src/cache.rs
+++ b/server/bleep/src/cache.rs
@@ -1,4 +1,5 @@
 use std::{
+    collections::HashSet,
     ops::Deref,
     path::PathBuf,
     sync::{Arc, RwLock},
@@ -6,9 +7,10 @@ use std::{
 };
 
 use qdrant_client::qdrant::{PointId, PointStruct};
+use rayon::prelude::ParallelIterator;
 use scc::hash_map::Entry;
 use sqlx::Sqlite;
-use tracing::{error, trace};
+use tracing::{error, info, trace, warn};
 use uuid::Uuid;
 
 use crate::{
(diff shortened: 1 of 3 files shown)


server/bleep/src/cache.rs
```diff
@@ -137,38 +260,20 @@ impl<'a> FileCache<'a> {
...
+        })
     }
 
     /// Commit the embed log, invoking the embedder if batch size is met.
     ///
     /// If `flush == true`, drain the log, send the entire batch to
     /// the embedder, and commit the results, disregarding the internal
     /// batch sizing.
-    pub async fn batched_process_embed_queue(&self, flush: bool) -> anyhow::Result<()> {
+    async fn batched_embed_or_flush_queue(&self, flush: bool) -> anyhow::Result<()> {
```
ggordonhall, line 276: Can we just merge this with the function below?

rsdy: It will make things fairly complicated, and that loop is quite dense as is. This is supposed to be an easy to grasp outline of how we generate embeddings, while the lower layer function is dealing with actually batching & making those embeddings.

PR #899, opened 2023-08-30
Add `DELETE /studio/:id`

calyptobai:


server/bleep/src/webserver.rs
```diff
@@ -79,6 +79,7 @@ pub async fn start(app: Application) -> anyhow::Result<()> {
         .route("/studio", get(studio::list))
         .route("/studio/:id", get(studio::get))
         .route("/studio/:id", patch(studio::patch))
+        .route("/studio/:id", delete(studio::delete))
```
nerdypepper, line 82: could compress these:
```suggestion
        .route("/studio/:id", get(studio::get).patch(studio::patch).delete(studio::delete))
```

nerdypepper (approved):

ggordonhall (approved):

anastasiya1155 (approved):

server/bleep/src/webserver.rs
calyptobai: Good point :slightly_smiling_face:

PR #901, opened 2023-08-30
Store incomplete conversations

calyptobai:
Now, conversations will persist in the database even if they do not properly complete. Failed conversations are still not persisted with this change, intentionally.

Closes BLO-1364

### DIFF
--- a/server/bleep/src/agent.rs
+++ b/server/bleep/src/agent.rs
@@ -1,9 +1,9 @@
-use std::time::Duration;
+use std::{sync::Arc, time::Duration};
 
 use anyhow::{anyhow, Context, Result};
-use futures::TryStreamExt;
+use futures::{Future, TryStreamExt};
 use tokio::sync::mpsc::Sender;
-use tracing::{debug, info, instrument};
+use tracing::{debug, error, info, instrument};
 
 use crate::{
     analytics::{EventData, QueryEvent},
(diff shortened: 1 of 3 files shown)


ggordonhall (changes requested):
We also want to add any incomplete, but cancelled, answers to the conversation history. So that the user could then follow up with a question like `list the other ones?` (referring to the partial answer in the previous step).

calyptobai:
> We also want to add any incomplete, but cancelled, answers to the conversation history. So that the user could then follow up with a question like `list the other ones?` (referring to the partial answer in the previous step).

To clarify, this is the intended behaviour of this change. The original PR wording is confusing, it should say "are cancelled" instead of "do not properly complete".

We add _cancelled_ answers, but **not** answers that _failed_ to complete.

ggordonhall:
> We add _cancelled_ answers, but **not** answers that _failed_ to complete.

I noticed that if I let bloop start generating an answer, then hit the button to cancel generation, the partial answer _does not_ end up in the history for subsequent questions.

calyptobai:
> I noticed that if I let bloop start generating an answer, then hit the button to cancel generation, the partial answer does not end up in the history for subsequent questions.

Ah I see the issue. The response was being stored on the back-end but we didn't display it properly on the front-end due to a missing conclusion. I've introduced some canned conclusions here, selected randomly from:

```
"The article wasn't completed. See what's available",
"Your article stopped before completion. Check out the available content",
"The content stopped generating early. Review the initial response"
```

ggordonhall (approved):

PR #903, opened 2023-08-30
filter related-files on currently selected ranges

nerdypepper:


nerdypepper:
closing this because the behaviour is unintuitive to end users.

PR #906, opened 2023-08-30
replace levenshtein fuzzy matching with skim fuzzy matcher

nerdypepper:
this should work more like how sublime text's search. generally, the
following changes in behavior are expected:

- better search results for camel and snake-case identifiers: a corpus
  containing `scope_resolution` is matched by `scores`
- ranks prefix matches higher: `res` matches both `resolution` and
  `progress`, but `resolution` will be ranked higher
- ranks file hits higher than directory hits
- ranks documents with hits near the file-stem higher than those with
  hits around the ancestor directories

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -493,6 +493,7 @@ dependencies = [
  "expect-test",
  "flume",
  "futures",
+ "fuzzy-matcher",
  "git-version",
  "gix",
  "histogram",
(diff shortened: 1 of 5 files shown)


nerdypepper:
tested on larger indexes (~14k docs), latency is reasonable (~100ms). we could do better by:
- introducing a new field in the index that breaks tokens on `/`
- introducing a new query type that does prefix matching 
- this query type would score documents with rules similar to fuzzy-matcher or nucleo-matcher's `match_paths`

ggordonhall:
> tested on larger indexes (~14k docs), latency is reasonable (~100ms). we could do better by:
> 
> * introducing a new field in the index that breaks tokens on `/`
> * introducing a new query type that does prefix matching
> * this query type would score documents with rules similar to fuzzy-matcher or nucleo-matcher's `match_paths`

Let's get this in for now and tackle these improvements at a later date.

ggordonhall (approved):

PR #907, opened 2023-08-30
Filter files imported to studio using GPT

stunningpixels:
When a repo conversation is imported into a studio project, the files should be filtered using GPT. This has been measured to increase the accuracy of the imported files.

If there is an error (GPT returns invalid JSON etc...) we fallback to importing all the context files.

### DIFF
--- a/server/bleep/src/webserver/studio.rs
+++ b/server/bleep/src/webserver/studio.rs
@@ -465,6 +465,7 @@ async fn extract_relevant_chunks(
 
     // Create an instance of the LLM gateway client
     let llm_gateway = llm_gateway::Client::new(&app.config.answer_api_url)
+        .model(LLM_GATEWAY_MODEL)
         .temperature(0.0)
         .bearer(answer_api_token);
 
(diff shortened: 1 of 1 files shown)


server/bleep/src/webserver/studio.rs
```diff
@@ -451,6 +451,74 @@ pub struct Import {
...
+
+    let answer_api_token = app
+        .answer_api_token()
+        .map_err(|e| Error::user(e).with_status(StatusCode::UNAUTHORIZED))?
+        .map(|s| s.expose_secret().clone());
+
+    // Create an instance of the LLM gateway client
+    let llm_gateway = llm_gateway::Client::new(&app.config.answer_api_url)
+        .temperature(0.0)
+        .bearer(answer_api_token);
```
calyptobai, line 469: We can specify the LLM model here to ensure it uses the same model as other studio calls.
```suggestion
    let llm_gateway = llm_gateway::Client::new(&app.config.answer_api_url)
        .model(LLM_GATEWAY_MODEL)
        .temperature(0.0)
        .bearer(answer_api_token);
```

calyptobai (approved):
LGTM. Left one nit for specifying a gateway model.

PR #912, opened 2023-08-31
Allow tweaking which files are indexed (EE)

rsdy:
There's some reorganization here to allow cleaner extension of filtering, but the core logic is unchanged.
Updating the file filter list will immediately trigger a reindexing, similarly to how branch filters work.

```
; curl 'http://127.0.0.1:7878/api/repos/indexed?repo=github.com%2Fbloopai%2Fbloop' \
-X 'PATCH' \
-H 'Content-Type: application/json' \
-H 'Accept: application/json, text/plain, */*' \
--data-binary '{"file_filter": {"rules": [{"include_file": "Cargo.toml"}, {"exclude_regex": ".*\\.toml"}]}}'

```

The following rules are recognized:
 * `include_file`
 * `exclude_file`
 * `include_regex`
 * `exclude_regex`
 
 Excludes always override includes.

If the file filters are specified, they override the built-in filterlist (unchanged) for files which they mention or the patterns match.

While `branch_filter` is additive, `file_filter` is absolute, and will delete the previous rule list when specified. Use `null` or omit from the structure to leave them untouched if only changing the `branch_filter` field.


server/bleep/src/indexes/file.rs
```diff
@@ -111,6 +119,7 @@ impl Indexable for File {
                     repo_disk_path: &repo.disk_path,
                     repo_ref: reporef.to_string(),
                     repo_name: &repo_name,
+                    file_filter: &file_filter,
```
ggordonhall, line 122: Do we need to pass the `file_filter` here, or can we just call this `indexed: bool` and bring this calculation:
```
let indexed = file_filter
            .is_allowed(relative_path)
            .unwrap_or_else(|| should_index(relative_path));
```           
 up into this method?

rsdy: We actually use this for cache key calculations as well, otherwise the forced reindexing logic won't work. It is better to keep this for this PR.

ggordonhall: Could we not have an `indexed: bool` field on `Workload` and calculate whether a file should be indexed in `index_repository` (before instantiating a `Workload`)?

rsdy: looking at this again, we don't only need a binary choice, since we ideally don't want to reindex everything with a new index key. so we need a ternary choice: denied, included, or not mentioned in the ruleset at all.

ggordonhall (approved):

PR #914, opened 2023-08-31
improve japanese translations

shoito:
Thank you for your great product.
I am a Japanese software engineer. Please let me suggest a more natural Japanese label.


shoito:
@ggordonhall Thanks for the merge.

PR #915, opened 2023-08-31
Add CRUD template functionality

stunningpixels:
GET /api/template
POST /api/template {name: string, content: string}
PATCH /api/template/:id  {name: string, content: string}
DELETE /api/template/:id

### DIFF
--- a/server/bleep/migrations/20230831165918_templates.sql
+++ b/server/bleep/migrations/20230831165918_templates.sql
@@ -4,3 +4,13 @@ CREATE TABLE templates (
     modified_at DATETIME NOT NULL DEFAULT (datetime('now')),
     content TEXT NOT NULL
 );
+
+INSERT INTO templates (
+  id,
+  name,
+  content
+) VALUES (
+  'DEFAULT_UNIT_TEST',
+  'Unit Test',
+  'Perform the following steps:\n\nStep 1: Describe the purpose of the code\nStep 2: Identify which functions would benefit from unit testing\nStep 3: Write a unit test for each of those functions'
+);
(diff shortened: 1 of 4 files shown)


server/bleep/migrations/20230831165918_templates.sql
```diff
@@ -0,0 +1,24 @@
+CREATE TABLE templates (
+    id TEXT NOT NULL,
```
calyptobai, line 2: Can we make this an `INTEGER PRIMARY KEY` instead? Sqlite will automatically give this column a default value if not specified, so we don't have to set the ID when inserting into the table. Integer keys are also easier to manage in general as they are `Copy`. In rust, this will be `i64`.

calyptobai (approved):
Left one comment about the ID type. I haven't tested this, but assuming it works, LGTM :+1:

server/bleep/migrations/20230831165918_templates.sql
```diff
@@ -0,0 +1,24 @@
...
+);
+
+INSERT INTO templates (
+  id,
+  name,
+  content
+) VALUES (
+  'DEFAULT_UNIT_TEST',
+  'Write a test',
+  'Perform the following steps:\n\nStep 1: Describe the purpose of the code\nStep 2: Identify which functions would benefit from unit testing\nStep 3: Write a unit test for each of those functions'
```
anastasiya1155, line 15: new lines get double escaped and return as \\n to the frontend which breaks rendering

anastasiya1155 (commented):
We should also add a template to refactor code [BLO-1533](https://linear.app/bloopai/issue/BLO-1533/template-to-refactor-a-section-of-code)

stunningpixels:
> We should also add a template to refactor code [BLO-1533](https://linear.app/bloopai/issue/BLO-1533/template-to-refactor-a-section-of-code)

Code review should cover this, generic refactor is difficult because the user will need to add guidance

PR #920, opened 2023-09-01
List studio repos and top file extension

anastasiya1155:
In the list of code studios 'GET /studio' endpoint added 2 new fields:
- `repos` containing repo refs used in this code studio, and
- `most_common_ext` containing the most used file extension based on a number of tokens in files. So if there are 3 files: foo.yml (10 tokens), bar.yml (100 tokens), and baz.rs (200 tokens) the response will contain `.rs`.

### DIFF
--- a/server/bleep/src/webserver/studio.rs
+++ b/server/bleep/src/webserver/studio.rs
@@ -1,4 +1,10 @@
-use std::{borrow::Cow, collections::HashMap, iter, mem, ops::Range, pin::Pin};
+use std::{
+    borrow::Cow,
+    collections::{HashMap, HashSet},
+    iter, mem,
+    ops::Range,
+    pin::Pin,
+};
 
 use anyhow::{Context, Result};
 use axum::{
@@ -57,7 +63,7 @@ pub struct ListItem {
     most_common_ext: String,
 }
 pub async fn list(app: Extension<Application>) -> webserver::Result<Json<Vec<ListItem>>> {
-    let studios = sqlx::query!("SELECT id, name, modified_at FROM studios")
+    let studios = sqlx::query!("SELECT id, name, modified_at, context FROM studios")
         .fetch_all(&*app.sql)
         .await
         .map_err(Error::internal)?;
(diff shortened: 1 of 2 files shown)


server/bleep/src/webserver/studio.rs
```diff
@@ -53,14 +53,50 @@ pub struct ListItem {
...
         .fetch_all(&*app.sql)
         .await
-        .map_err(Error::internal)
-        .map(Json)
+        .map_err(Error::internal)?;
+
+    let mut list_items = Vec::new();
+
+    for studio in studios {
+        let context_json: String = sqlx::query!("SELECT context FROM studios WHERE id = ?", studio.id)
```
anastasiya1155, line 68: I wanted to avoid this additional sqlx call for each studio, but putting context into select on line 60 just returns an error
```
error: failed to find data for query SELECT id, name, modified_at, context FROM studios
  --> server/bleep/src/webserver/studio.rs:68:35
   |
68 |       let studios: Vec<StudioRow> = sqlx::query_as!(
   |  ___________________________________^
69 | |         StudioRow,
70 | |         "SELECT id, name, modified_at, context FROM studios"
71 | |     )
   | |_____^
   |
   = note: this error originates in the macro `$crate::sqlx_macros::expand_query` which comes from the expansion of the macro `sqlx::query_as` (in Nightly builds, run with -Z macro-backtrace for more info)
```

calyptobai: We should swap this back to a single SQL call. The reason it was giving that error is that you have to use the `sqlx` CLI to update the query info JSON. `sqlx` actually loads the database and examines the query in order to do this.

server/bleep/src/webserver/studio.rs
```diff
@@ -53,14 +53,56 @@ pub struct ListItem {
...
+        let mut repos: Vec<String> = context.iter().map(|file| file.repo.name.clone()).collect();
+        repos.sort();
+        repos.dedup();
+
+        let mut ext_tokens = HashMap::new();
+        for file in &context {
+            let ext = file.path.split('.').last().unwrap_or("");
+            let tokens = token_counts((*app).clone(), &[], &[file.clone()]).await?;
+            *ext_tokens.entry(ext.to_string()).or_insert(0) += tokens.total;
+        }
```
nerdypepper, line 87: `token_counts` can be used to calculate all token counts at once. this loop might be replaced with:

```rust
token_counts(_, _, &context).await?
    .per_file
    .zip(context.iter().map(|file| file.path.split('.').last().unwrap_or_default()))
    .fold(HashMap::new(), |ext_tokens, (count, ext)| {
        *ext_tokens.entry(ext.to_string()).or_insert(0) += count;
        ext_tokens
    });
```

PR #921, opened 2023-09-01
Use msgpack for embedder server exchange

rsdy:


server/bleep/src/ee/embedder.rs
```diff
@@ -17,20 +18,23 @@ impl RemoteEmbedder {
         let url = url.join("encode")?;
         Ok(Self {
             url,
-            session: reqwest::Client::builder().gzip(true).build()?,
+            session: reqwest::Client::builder().gzip(false).build()?,
```
rsdy, line 21: Do we need to disable accepting gzip?

PR #926, opened 2023-09-01
Implement snapshot-based studio history

calyptobai:
This PR adds "snapshots" to studios. By default, using the routes as before changes nothing, but new "snapshots" are created during `/import` and `/generate` calls. When a snapshot is taken, the current state is saved and duplicated. Most routes are by-default unaware of this, here are the complete changes:

- `GET /studio` retrieves the full list of studios, each titled and dated by their latest snapshot
- `GET /studio/:id` returns the "latest" snapshot, but optionally takes `?snapshot_id=...`
- `PATCH /studio/:id` operates on the "latest" snapshot, but optionally takes `?snapshot_id=...`
- `DELETE /studio/:id` works as before, but also deletes **all** associated snapshots
- `POST /studio/:id/import` works as before, but now implicitly creates a new snapshot on successful import
- `GET /studio/:id/generate` works as before, but now implicitly creates a new snapshot on successful generation

To manage these snapshots, we add 2 new routes:

- `GET /studio/:studio_id/snapshots`
  - This will return a list of snapshots, which look like:
    ```ts
    {
      id: number,
      modified_at: date string,
      context: [...],
      messages: [...]
    }
    ```
- `DELETE /studio/:studio_id/snapshots/:snapshot_id`
  - This deletes the snapshot

Closes BLO-1535

### DIFF
--- a/server/bleep/migrations/20230831184906_code_studio_history.sql
+++ b/server/bleep/migrations/20230831184906_code_studio_history.sql
@@ -1,0 +1,31 @@
+ALTER TABLE studios RENAME TO studios_old;
+
+CREATE TABLE studios (
+    id BLOB NOT NULL,
+    name TEXT NOT NULL,
+
+    PRIMARY KEY (id)
+);
+
+CREATE TABLE studio_snapshots (
+    id INTEGER PRIMARY KEY,
+    studio_id BLOB NOT NULL,
+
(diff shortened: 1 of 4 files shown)


calyptobai:
Reviewed offline, merging as per request from @stunningpixels

PR #929, opened 2023-09-04
Only call register after github login

rsdy:


anastasiya1155 (commented):
will this work both on cloud and in the app?

rsdy:
Yes, cloud seems to have another call in place in the right spot.

anastasiya1155 (approved):

nerdypepper (approved):

PR #931, opened 2023-09-05
Take `user_id` into account for studio routes

calyptobai:
This adds a new `user_id` column to the `studios` table and ***will thus drop all existing studios to avoid a table with null IDs***.


calyptobai:
Merging without review as per offline discussion with @stunningpixels

PR #932, opened 2023-09-06
offload embedding to the gpu if possible

nerdypepper:
introduces gpu indexing under a feature flag. summary of feature flags introduced in this PR:
- `onnx`: this flag is the default, uses `onnx` on cpu to produce embeddings
- `ggml` (or `metal`): this flag can be turned on if `--no-default-features` is used, uses `ggml` to run the embed step on the gpu

--- 

the core logic is introduced in the `llm` crate, but `semantic::embedder::gpu::LocalEmbedder` contains the code required to interface with `llm` and therefore `ggml`. the old cpu emebdder is now stashed under `semantic::embedder::cpu::LocalEmbedder`.


server/bleep/Cargo.toml
```diff
@@ -134,6 +138,9 @@ lazy-regex = "3.0.2"
 quick-xml = { version = "0.29.0", features = ["serialize"] }
 jsonwebtokens-cognito = "0.1.1"
 
+# llm = { path = "/home/np/bloop/experimental/llm/crates/llm", default-features = false, features = ["bert"] }
+llm = { git = "https://github.com/nerdypepper/llm", branch = "batch-metal-kernels", default-features = false, features = ["bert"], optional = true }
```
ggordonhall, line 142: Can you move this into the bloop org?

server/bleep/src/config.rs
```diff
@@ -421,5 +421,6 @@ fn default_max_chunk_tokens() -> usize {
 }
 
 fn interactive_batch_size() -> NonZeroUsize {
-    NonZeroUsize::new(1).unwrap()
+    let batch_size = if cfg!(feature = "metal") { 5 } else { 1 };
```
ggordonhall, line 424: How was 5 chosen? There is some evidence that batch sizes that are powers of 2 are more efficient. Did we benchmark 4?

server/bleep/src/semantic.rs
```diff
@@ -34,12 +34,11 @@ pub enum SemanticError {
     #[error("Qdrant initialization failed. Is Qdrant running on `qdrant-url`?")]
     QdrantInitializationError,
 
-    #[error("ONNX runtime error")]
-    OnnxRuntimeError {
-        #[from]
-        error: ort::OrtError,
-    },
-
+    // #[error("ONNX runtime error")]
```
ggordonhall, line 37: Could we stick this behind a feature flag?

server/bleep/src/semantic.rs
```diff
@@ -34,12 +34,11 @@ pub enum SemanticError {
...
+    // OnnxRuntimeError {
+    //     #[from]
+    //     error: ort::OrtError,
+    // },
```
nerdypepper, line 41: uncomment this

server/bleep/src/semantic/embedder.rs
```diff
@@ -65,89 +59,261 @@ pub struct EmbedChunk {
...
+                model_params,
+                llm::load_progress_callback_stdout,
+            )?;
+
+            // TODO: this can be parameterized
+            //
+            // the lower this number, the more time we might spend waiting to run an embedding.
+            // the higher this number, the more vram we use, currently we use ~2G per session. this
+            // can be fixed by disabling scratch buffers in ggml, bert has no use for this.
+            let session_count = 3;
```
nerdypepper, line 212: the difference in inference speed with a lower session count is minor, we see a ~6% speedup when indexing bloop with `3` instead of `1`.

server/bleep/src/semantic/embedder.rs
```diff
@@ -65,89 +59,261 @@ pub struct EmbedChunk {
...
-        &self.tokenizer
+    // InferenceSession is explicitly not Sync because it uses ggml::Tensor internally,
+    // Bert does not make use of these tensors however
+    unsafe impl Sync for LocalEmbedder {}
+
+    impl LocalEmbedder {
+        pub fn new(model_dir: &Path) -> anyhow::Result<Self> {
+            let mut model_params = llm::ModelParameters::default();
+
+            if cfg!(feature = "metal") {
```
ggordonhall, line 192: This check is redundant. Can just hardcode `use_gpu = true` for now.

server/bleep/src/semantic/embedder.rs
```diff
@@ -65,89 +59,261 @@ pub struct EmbedChunk {
...
+                all_logits: None,
+                embeddings: Some(Vec::new()),
+            };
+            let vocab = self.model.tokenizer();
+            let beginning_of_sentence = true;
+            let query_token_ids = log
+                .iter()
+                .map(|sequence| {
+                    vocab
+                        .tokenize(&sequence, beginning_of_sentence)
```
ggordonhall, line 284: At some point we should amend `llm` to expose parallel tokenization: https://docs.rs/tokenizers/latest/tokenizers/tokenizer/struct.Tokenizer.html

server/bleep/src/semantic/embedder.rs
```diff
@@ -65,89 +59,261 @@ pub struct EmbedChunk {
...
+                    if let Ok(mut session) = s.try_lock() {
+                        self.model.batch_evaluate(
+                            &mut session,
+                            &query_token_ids,
+                            &mut output_request,
+                        );
+                        let embedding: Vec<Vec<f32>> = output_request
+                            .embeddings
+                            .unwrap()
+                            .chunks(384)
```
ggordonhall, line 304: This should be set to `EMBEDDING_DIM`: https://github.com/BloopAI/bloop/blob/7e2cf6bcb207931261ddc962090dda852ab997c4/server/bleep/src/semantic/schema.rs#L12

ggordonhall (approved):

server/bleep/src/config.rs
nerdypepper: 5 was the smallest number that did not run out of context-memory. will test this again with 4.

server/bleep/src/semantic/embedder.rs
```diff
@@ -65,89 +59,258 @@ pub struct EmbedChunk {
...
-    session: ort::Session,
-    tokenizer: Tokenizer,
-}
+#[cfg(not(feature = "metal"))]
+pub use cpu::LocalEmbedder;
+#[cfg(feature = "metal")]
+pub use gpu::LocalEmbedder;
+
+#[cfg(not(feature = "metal"))]
+mod cpu {
```
rsdy, line 73: it would be nice to move these to actually separate files.

server/bleep/src/semantic/embedder.rs
```diff
@@ -65,89 +59,258 @@ pub struct EmbedChunk {
...
+        }
+
+        async fn batch_embed(&self, log: Vec<&str>) -> anyhow::Result<Vec<Embedding>> {
+            let mut output_request = llm::OutputRequest {
+                all_logits: None,
+                embeddings: Some(Vec::new()),
+            };
+            let vocab = self.model.tokenizer();
+            let beginning_of_sentence = true;
+            let query_token_ids = log
```
rsdy, line 277: This seems repeated from `embed`. Is it possible to express `batch_embed` in terms of `embed`, or parameterize the common bits to make it easy to spot the differences?

rsdy (changes requested):

nerdypepper:
superseeded by #1103

PR #933, opened 2023-09-07
Add proxy billing routes

calyptobai:
Depends on answer-api#54

Here, we add 2 new routes:

- `GET /api/quota`: This will return an object like `{ paid: boolean, used: string (a stringified number) }`
- `GET /api/quota/create-checkout-session`: This will create a checkout session if the user is new, or if they have already purchased a plan, return a link to adjust the plan. The returned object looks like `{ url: string }`, with a Stripe URL.

Additionally, each call to `/studio/:id/generate` will be quota-limited, such that after hitting 10 requests in one day (as configured currently), future calls will fail. The front-end can ensure we don't hit an error here by disabling the `Generate` button when the `used` value is below 10.

Note that the Answer API must be quota-enabled for this to work. Instructions are listed in the relevant PR.

### DIFF
--- a/server/bleep/src/llm_gateway.rs
+++ b/server/bleep/src/llm_gateway.rs
@@ -88,6 +88,7 @@ pub mod api {
         #[serde(default)]
         pub extra_stop_sequences: Vec<String>,
         pub session_reference_id: Option<String>,
+        pub quota_gated: bool,
     }
 
     #[derive(Debug, Copy, Clone, serde::Serialize, serde::Deserialize)]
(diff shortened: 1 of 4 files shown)


anastasiya1155:
I just understood that we need to return the daily limit of requests and it shouldn't be hardcoded on the frontend, so I will change the response from lambda to contain the 'allowed' field set to 10. I will also make 'used' field a number.

calyptobai:
> I just understood that we need to return the daily limit of requests and it shouldn't be hardcoded on the frontend, so I will change the response from lambda to contain the 'allowed' field set to 10. I will also make 'used' field a number.

I've adjusted this now, we return `allowed` and `used` is also a number. `paid` has been also renamed to `upgraded`.

calyptobai:
Merging as per offline discussion with @stunningpixels

PR #935, opened 2023-09-07
Studio: get file tokens

anastasiya1155:
With the latest change on the FE, we no longer save a file to context as soon as it is selected, nor when the line ranges change.  Now we save changes on submit. This leads to an issue that the token counter cannot show the current number of tokens when the user adds a file to context or changes ranges. This PR adds a separate endpoint to only get a number of tokens for a given file (and optionally ranges). Note that this endpoint is a POST, although it would make more sense to make it a GET, but passing an array of arrays in query params turned out to be extremely tricky.

### DIFF
--- a/server/bleep/src/webserver/studio.rs
+++ b/server/bleep/src/webserver/studio.rs
@@ -501,6 +501,61 @@ async fn token_counts(
     })
 }
 
+#[derive(serde::Deserialize)]
+pub struct GetTokenCount {
+    pub path: String,
+    pub repo: RepoRef,
+    pub branch: Option<String>,
+    pub ranges: Option<Vec<Range<usize>>>,
+}
+
+pub async fn get_file_token_count(
+    app: Extension<Application>,
+    Json(params): Json<GetTokenCount>,
+) -> webserver::Result<Json<usize>> {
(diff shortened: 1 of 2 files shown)


nerdypepper (approved):

server/bleep/src/webserver/studio.rs
```diff
@@ -501,6 +501,61 @@ async fn token_counts(
...
+                .iter()
+                .copied()
+                .skip(range.start)
+                .take(range.end - range.start)
+                .collect::<Vec<_>>()
+                .join("\n");
+
+            token_count += core_bpe.encode_ordinary(&chunk).len();
+        }
+    }
```
calyptobai, line 554: Can we extract this into a function? We can use the function [above](https://github.com/BloopAI/bloop/pull/935/files#diff-663c689783c1e5c15f880b3602f7455c2dbc19b71245624756e4d8fe11fcb689R446-R466) as well, to avoid duplication of this token counting logic.

anastasiya1155: done

calyptobai (approved):
LGTM :slightly_smiling_face:

PR #938, opened 2023-09-07
Change studio ID type from UUID to integer

calyptobai:
**Warning**: This will drop the current studio tables. Given that a core ID was changed and the feature is not yet released, we just clear them instead of a more complex migration.

All uses of UUIDs to manage the `studios` table have been switched to a plain `INTEGER` as a primary key.

Closes BLO-1575


calyptobai:
Tested by @stunningpixels, merging

PR #951, opened 2023-09-13
Auto-generate studio names

calyptobai:
Now, studio names are given a default name of `New Studio` if a name is not supplied at creation time. These studios will internally have a name of `NULL`, which will then attempt to be auto-populated in a call to `/studio/:id/generate`.

Note, the front-end currently still requires a studio name to be specified. To test this behaviour, you can set a studio's name to `NULL` with a sqlite database editor, and generate a response which will in turn populate the studio name.

Closes BLO-1505

### DIFF
--- a/server/bleep/src/webserver/studio.rs
+++ b/server/bleep/src/webserver/studio.rs
@@ -16,8 +16,7 @@ use chrono::NaiveDateTime;
 use futures::{pin_mut, stream, StreamExt, TryStreamExt};
 use rayon::prelude::*;
 use reqwest::StatusCode;
-use secrecy::ExposeSecret;
-use tracing::error;
+use tracing::{debug, error};
 use uuid::Uuid;
 
 use super::{middleware::User, Error};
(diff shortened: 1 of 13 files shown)


server/bleep/src/webserver/studio.rs
```diff
@@ -680,6 +681,62 @@ async fn generate_llm_context(app: Application, context: Vec<ContextFile>) -> Re
...
+        .map_err(|e| Error::user(e).with_status(StatusCode::UNAUTHORIZED))?
+        .model(LLM_GATEWAY_MODEL)
+        .temperature(0.0);
+
+    let messages = &[llm_gateway::api::Message::system(
+        &prompts::studio_name_prompt(&context_json, &messages_json),
+    )];
+
+    let name = llm_gateway
+        .chat(messages, None)
```
nerdypepper, line 726: can we use 3.5-turbo heres/

nerdypepper (approved):

ggordonhall (approved):
LGTM. Would be cool in future to add a mechanism to update studio names as the conversation progresses.

server/bleep/src/webserver/studio.rs
ggordonhall: Have tested and GPT-3.5 turbo seems sufficient

PR #968, opened 2023-09-19
Fix security issue with actor validation

corest:


.github/workflows/build-on-pr-command.yml
```diff
@@ -25,13 +25,16 @@ jobs:
...
             -H "X-GitHub-Api-Version: 2022-11-28" \
-            https://api.github.com/orgs/BloopAI/members | jq -r ".[] .login" | grep ${ACTOR}; then
-            echo "is-member=true" >> $GITHUB_OUTPUT
-          fi
+            https://api.github.com/orgs/BloopAI/members | jq -r ".[] .login")
+          for member in $(echo ${members}); do 
+            if [[ $member = $ACTOR ]]; then 
+              echo "is-member=true" >> $GITHUB_OUTPUT
+            fi
+          done
```
nerdypepper, line 37:
```suggestion
          for member in "${members[@]}"; do 
            [[ "$member" == "$ACTOR" ]] && echo "is-member=true" >> $GITHUB_OUTPUT
          done
```
or we could change `grep` to use `-F` and `-w`.

corest: I like to be explicit. Also this piece doesn't work in my shell

PR #983, opened 2023-09-26
Make use of our tokio-enabled rayon thread pool explicit

calyptobai:
Here, we use an explicit `rayon` pool when queuing files. It was previously possible to break our `rayon` usage if a global pool was initialized before our background executor setup. In this case, `process_embedding_queue` would call `tokio::runtime::Handle::current` in a rayon thread that does not have re-entrant behaviour with `tokio`, which results in immediate failure.

This was observed with related work that utilized `rayon` in unrelated parts of the application, which broke worker queuing.

### DIFF
--- a/server/bleep/src/background.rs
+++ b/server/bleep/src/background.rs
@@ -1,3 +1,5 @@
+use once_cell::sync::OnceCell;
+use rayon::ThreadPool;
 use thread_priority::ThreadBuilderExt;
 use tokio::sync::Semaphore;
 use tracing::{debug, info};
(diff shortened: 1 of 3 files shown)


rsdy (changes requested):
In my tests this has increased indexing time from 36s to 99s. I suspect what's happening is that `sysinfo` in #984 is interfering with our global rayon configuration.

Dropping rayon here effectively disables a large part of per-file parallelism in the indexing pipeline, which is why we see the perf hit for indexing. We may need to consider alternative approaches to resolve all of this, seems like a right mess to me.

rsdy (approved):

PR #987, opened 2023-09-26
Rework index initialisation logic

rsdy:
This PR's aim is to make the qdrant dependency mandatory. In practice, we always want semantic indexing, paying attention to this every time we use this subsystem is not super productive.

In addition, steps are taken to reduce some of the dependency mess between tantivy & semantic indexes. `FileCache` is now hierarchically adjacent to the `File` indexer, which opened the door to removing a bunch of database dependnencies for that layer. The `FileCache` is now managed by `SyncHandle`, and passed on as a reference.

The initialization flow is also somewhat clearer, and at a high level is contained fully in `lib.rs` for easier reading.

### DIFF
--- a/server/bleep/src/cache.rs
+++ b/server/bleep/src/cache.rs
@@ -67,7 +67,8 @@ impl<T> From<T> for FreshValue<T> {
 /// representative at a single point in time
 pub struct FileCacheSnapshot<'a> {
     snapshot: Arc<scc::HashMap<CacheKeys, FreshValue<()>>>,
-    parent: &'a FileCache<'a>,
+    parent: &'a FileCache,
+    reporef: &'a RepoRef,
 }
 
 /// CacheKeys unifies the different keys to different databases.
@@ -104,7 +105,7 @@ impl CacheKeys {
 }
 
 impl<'a> FileCacheSnapshot<'a> {
-    pub(crate) fn parent(&'a self) -> &'a FileCache<'a> {
+    pub(crate) fn parent(&'a self) -> &'a FileCache {
         self.parent
     }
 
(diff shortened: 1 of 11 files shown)


server/bleep/src/cache.rs
```diff
@@ -390,20 +393,18 @@ impl<'a> FileCache<'a> {
         &self,
         cache_keys: &CacheKeys,
         repo_name: &str,
-        repo_ref: &str,
+        repo_ref: &RepoRef,
```
ggordonhall, line 396: Can update the comment on L390

server/bleep/src/indexes/file.rs
```diff
@@ -42,10 +39,11 @@ use crate::{
 struct Workload<'a> {
     cache: &'a FileCacheSnapshot<'a>,
     file_filter: &'a FileFilter,
+    repo_ref: &'a RepoRef,
     repo_disk_path: &'a Path,
     repo_name: &'a str,
     repo_metadata: &'a RepoMetadata,
-    repo_ref: String,
+    repo_ref_str: String,
```
ggordonhall, line 46: Confusing having both `repo_ref` and `repo_ref_str` here. Why don't we just have `repo_ref` and transform into a `String` when needed.

ggordonhall (approved):
Nice! The initialisation logic is now much easier to follow along.

server/bleep/src/indexes/file.rs
```diff
@@ -117,9 +114,10 @@ impl Indexable for File {
 
                 let workload = Workload {
                     repo_disk_path: &repo.disk_path,
-                    repo_ref: reporef.to_string(),
+                    repo_ref_str: reporef.to_string(),
                     repo_name: &repo_name,
                     file_filter: &file_filter,
+                    repo_ref: reporef,
```
nerdypepper, line 120: do we need these both? `workload.repo_ref_str` is deriavable from `workload.repo_ref.to_string()` no?

nerdypepper (approved):

PR #989, opened 2023-09-27
Force agent to answer after maximum turn count

calyptobai:
Closes BLO-1650

### DIFF
--- a/server/bleep/src/agent.rs
+++ b/server/bleep/src/agent.rs
@@ -21,6 +21,9 @@ use crate::{
 
 use self::exchange::{Exchange, SearchStep, Update};
 
+/// The maximum number of steps the agent will take before forcing an answer.
+const MAX_STEPS: usize = 10;
+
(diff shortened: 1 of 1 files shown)


server/bleep/src/agent.rs
```diff
@@ -181,6 +184,12 @@ impl Agent {
             Action::Proc { query, paths } => self.process_files(query, paths).await?,
         };
 
+        if self.exchanges.len() >= MAX_STEPS {
+            return Ok(Some(Action::Answer {
+                paths: self.paths().enumerate().map(|(i, _)| i).collect(),
```
ggordonhall, line 189: I haven't tested the answer quality, but it might be better if we get bloop itself to choose which paths to answer w.r.t.

E.g. If we exceed `MAX_STEPS` we force the agent to call the `none` function: https://platform.openai.com/docs/api-reference/chat/create#function_call

PR #991, opened 2023-09-27
Agent efficiency improvements

ggordonhall:
Changes to reduce the number of GPT-calls per-query, lowering query latency:
- Always run a code search on initial query
- Replace proc GPT calls with path-constrained semantic code search
- Prompt agent to look for README files for generic queries
- Only run HyDE if the unexpanded query returns no results


server/bleep/src/agent.rs
```diff
@@ -168,6 +169,28 @@ impl Agent {
         match &action {
             Action::Query(s) => {
                 self.track_query(EventData::input_stage("query").with_payload("q", s));
+
+                // Always make a code search for the user query on the first exchange
+                if self.exchanges.len() == 1 {
+                    // Extract keywords from the query
+                    let keywords = {
+                        let sw = rake::StopWords::from_file("./stopwords.txt").unwrap();
```
nerdypepper, line 177: this could be statically initialized

nerdypepper (approved):

PR #996, opened 2023-09-29
Reindex local files on file change

calyptobai:
Now, local files are re-indexed whenever a file is changed. Files are added to the index under the current repository branch.

Local repositories will now also be indexed when new files are added.

Closes BLO-1640

### DIFF
--- a/server/bleep/src/repo/iterator/fs.rs
+++ b/server/bleep/src/repo/iterator/fs.rs
@@ -8,10 +8,13 @@ use std::path::{Path, PathBuf};
 
 pub struct FileWalker {
     file_list: Vec<PathBuf>,
+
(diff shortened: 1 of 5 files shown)


server/bleep/src/repo/iterator/fs.rs
```diff
@@ -8,10 +8,13 @@ use std::path::{Path, PathBuf};
 
 pub struct FileWalker {
     file_list: Vec<PathBuf>,
+
+    /// Branch to index files as part of.
+    branch: Option<String>,
```
rsdy, line 13: What happens if this value is `None`?

ggordonhall (approved):

PR #1023, opened 2023-10-04
Fix inner markdown

calyptobai:
In the agent, this is resolved by correcting the XML to backtick conversion, such that outer backtick counts will always be at least one greater than the count of any backtick groups on the start of a line, within the code snippet.

In studio, this is resolved via prompting to ensure that markdown blocks are generated with 4 backticks instead of the standard 3.

Closes BLO-1652


ggordonhall (approved):

Q755:
@ggordonhall

PR #1024, opened 2023-10-06
Use accurate token counts

calyptobai:
The `total` token count is now based upon an identical format conversation sent to the LLM during response generation.

This results in counts that should be accurate, and prevent token limit errors entirely.


anastasiya1155:
it is much closer now, but I still see some weird discrepancy
Screenshot 2023-10-10 at 12 35 07
Screenshot 2023-10-10 at 12 35 17

ggordonhall (approved):

PR #1048, opened 2023-10-13
Don't use credentials to fetch public repos

rsdy:
This can occasionally cause issues with certain organizations which enforce a SAML policy regime on their staff.

Even if the repository is public, access by a 3rd party app can be denied. Therefore for all public repos, bleep will not use credentials.

### DIFF
--- a/server/bleep/src/remotes/github.rs
+++ b/server/bleep/src/remotes/github.rs
@@ -130,10 +130,13 @@ impl Auth {
             Ok(details) => details,
         };
 
-        Ok(repo
-            .private
-            .map(|private| if private { Some(self.git_cred()) } else { None })
-            .unwrap_or_else(|| Some(self.git_cred())))
+        Ok(match repo.private {
+            // No credentials for public repos
+            Some(false) => None,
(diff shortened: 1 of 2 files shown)


server/bleep/src/remotes/github.rs
```diff
@@ -115,18 +115,25 @@ impl Auth {
...
             // Octocrab seems to treat GitHub application-layer errors as higher priority
-            Err(octocrab::Error::Http { .. }) => Err(RemoteError::PermissionDenied),
-            _ => Ok(response.map(|_| ())?),
-        }
+            Err(octocrab::Error::Http { .. }) => return Err(RemoteError::PermissionDenied),
+            Err(err) => return Err(err)?,
+            Ok(details) => details,
+        };
+
+        Ok(repo
```
ggordonhall, line 133: A `map_or` would be neater here:

```rust
    let creds = repo.private.map_or(Some(self.git_cred()), |private| {
        if private {
            Some(self.git_cred())
        } else {
            None
        }
    });

    Ok(creds)
```

ggordonhall (approved):
Looks good. Just one nit.

PR #1057, opened 2023-10-17
Catch unwind file indexing iterator

rsdy:
If we allow the file processing to crash, it will gradually eliminate all Rayon worker threads. Log these errors, but don't let Rayon threads die.

### DIFF
--- a/server/bleep/src/indexes/file.rs
+++ b/server/bleep/src/indexes/file.rs
@@ -1,5 +1,6 @@
 use std::{
     collections::{HashMap, HashSet},
+    panic::AssertUnwindSafe,
     path::{Path, PathBuf},
     sync::atomic::{AtomicU64, Ordering},
 };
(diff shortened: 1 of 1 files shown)


server/bleep/src/periodic/remotes.rs
```diff
@@ -48,11 +48,10 @@ async fn sleep_systime(duration: Duration) {
         return tokio::time::sleep(duration).await;
     }
 
-    let start = SystemTime::now();
-
+    let start = Instant::now();
```
calyptobai, line 51: As discussed last week, we want to rely on `SystemTime` here as time tracked by `Instant` will not pass on some systems when the OS is suspended.

calyptobai (commented):
Was this superseded by #1052?

rsdy:
@calyptobai I didn't notice how dirty this patchset was. In reality, this is supposed to be 1 patch. Updated now.

calyptobai (approved):
LGTM :+1:

One thing: I think the commit message is slightly misleading here as it says "but don't let Rayon threads die" - this might be outdated text?

PR #1063, opened 2023-10-20
Upload symbols to Sentry

ggordonhall:
Upload debug symbols to Sentry for MacOS builds so that we get more informative stack-traces where a call-site has been inlined.

### DIFF
--- a/.github/workflows/tauri-release.yml
+++ b/.github/workflows/tauri-release.yml
@@ -152,6 +152,7 @@ jobs:
         if: matrix.name == 'macos-11'
         run: |
           sentry-cli debug-files upload \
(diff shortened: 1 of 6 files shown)


rsdy (approved):
Did anything change in `bleep/Cargo.toml`? It's a bit hard to see if it's just a reorg from the diff.

ggordonhall:
> Did anything change in `bleep/Cargo.toml`? It's a bit hard to see if it's just a reorg from the diff.

Only one thing. I removed `git-version = "0.3.5"` from the `dev-dependencies`. It's already in `dependencies`.

calyptobai (approved):

PR #1067, opened 2023-10-24
Refactor credentials handling

rsdy:
There are a few things here:

 * put all background jobs to their dedicated single-threaded executor, so we only debug one at a time instead of a complex multi-threaded soup (and thus avoid [strange tokio bugs](https://github.com/tokio-rs/tokio/issues/4730))
 * eliminate use of `flume` to signal updated credentials, this isn't necessary anymore, because of the next item
 * centralize the llm constructor & some user management so we can make sure we update the creds whenever they expire
 * get rid of `futures::join_all` everywhere in favour of `tokio::task::JoinSet`

### DIFF
--- a/server/bleep/src/periodic.rs
+++ b/server/bleep/src/periodic.rs
@@ -2,7 +2,7 @@ mod logrotate;
 mod remotes;
 
 use logrotate::*;
-pub use remotes::*;
+pub(crate) use remotes::*;
 
 use crate::Application;
 
--- a/server/bleep/src/periodic.rs
+++ b/server/bleep/src/periodic.rs
@@ -1,5 +1,53 @@
 mod logrotate;
 mod remotes;
 
-pub(crate) use logrotate::*;
-pub(crate) use remotes::*;
+use logrotate::*;
+use remotes::*;
+
+use crate::Application;
+
+pub(crate) fn start_background_jobs(app: Application) {
+    if !app.env.is_cloud_instance() {
+        let app = app.clone();
+        std::thread::spawn(move || {
+            tokio::runtime::Builder::new_current_thread()
+                .enable_all()
+                .build()
+                .expect("failed to start background jobs")
+                .block_on(clear_disk_logs(app.clone()));
+        });
+    }
+
+    {
+        let app = app.clone();
+        std::thread::spawn(move || {
+            tokio::runtime::Builder::new_current_thread()
+                .enable_all()
+                .build()
+                .expect("failed to start background jobs")
+                .block_on(sync_github_status(app.clone()));
+        });
+    }
+
+    {
+        let app = app.clone();
+        std::thread::spawn(move || {
+            tokio::runtime::Builder::new_current_thread()
+                .enable_all()
+                .build()
+                .expect("failed to start background jobs")
+                .block_on(check_repo_updates(app.clone()));
+        });
+    }
+
+    {
+        let app = app.clone();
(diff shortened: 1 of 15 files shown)


server/bleep/src/periodic/remotes.rs
```diff
@@ -1,7 +1,7 @@
 use std::{
```
ggordonhall, line 1: Maybe worth using `tokio::time::timeout` rather than the `select!` block on line 313 of this file?

server/bleep/src/webserver/middleware.rs
```diff
@@ -16,34 +16,73 @@ use tracing::error;
...
+    pub(crate) fn org_name(&self) -> Option<&str> {
+        let User::Cloud { org_name, .. } = self else {
             return None;
         };
 
-        Some(login)
+        Some(org_name.as_ref())
     }
 
     pub(crate) fn github(&self) -> Option<octocrab::Octocrab> {
```
ggordonhall, line 51: Can you rename this `github_client`?

server/bleep/src/webserver/middleware.rs
```diff
@@ -16,34 +16,73 @@ use tracing::error;
...
+        let crab = match self {
+            User::Unknown => return None,
+            User::Desktop { crab, .. } => crab,
+            User::Cloud { crab, .. } => crab,
         };
 
         crab().ok()
     }
 
+    pub(crate) fn api_token(&self) -> Option<&str> {
```
ggordonhall, line 61: Calling this `access_token` would make the naming consistent across the backend.

server/bleep/src/webserver/middleware.rs
```diff
@@ -16,34 +16,73 @@ use tracing::error;
...
+        login: String,
         #[serde(skip)]
         crab: Arc<dyn Fn() -> anyhow::Result<octocrab::Octocrab> + Send + Sync>,
     },
 }
 
 impl User {
-    pub(crate) fn login(&self) -> Option<&str> {
-        let User::Authenticated { login, .. } = self else {
+    pub fn login(&self) -> Option<&str> {
```
ggordonhall, line 35: Can we rename this `id` since all it does is retrieve the user id.

ggordonhall (approved):
In general this makes sense. This logic is easier to follow now too.

server/bleep/src/webserver/middleware.rs
rsdy: Github calls it `login`, and it's not an opaque symbol, so i'd steer away from `id`. It's also not any internal user id, but the user's visible login name as per Github.

calyptobai: What about `username`? I've also found `login` confusing in the past before.

server/bleep/src/periodic.rs
```diff
@@ -1,5 +1,53 @@
...
+
+    {
+        let app = app.clone();
+        std::thread::spawn(move || {
+            tokio::runtime::Builder::new_current_thread()
+                .enable_all()
+                .build()
+                .expect("failed to start background jobs")
+                .block_on(log_and_branch_rotate(app.clone()));
+        });
```
calyptobai, line 51: Can we extract the duplicate logic here to a function like `start_background_job`?

calyptobai (approved):
I have left some minor comments, otherwise this LGTM

PR #1078, opened 2023-10-25
Recover from network errors in `poll_for_oauth_token`

ggordonhall:
This allows bloop to recover from any network errors that we hit when fetching the OAuth token. Previously we would exit the loop and authorisation would fail.


rsdy (approved):
This makes sense, I think I only considered upstream errors (5xx) during this, not arbitrary network errors.

PR #1103, opened 2023-11-01
offload embeddings to gpu

nerdypepper:


apps/desktop/src-tauri/tauri.conf.json
```diff
@@ -60,14 +60,13 @@
...
         "frameworks": [
-          "frameworks/libonnxruntime.dylib"
+            "frameworks/libonnxruntime.dylib"
         ],
         "providerShortName": null,
         "signingIdentity": null
       },
       "resources": [
         "model/*",
-        "dylibs/*",
```
rsdy, line 70: don't we still need this for linux builds?

rsdy (approved):

ggordonhall (approved):

apps/desktop/src-tauri/tauri.conf.json
nerdypepper: have re-added this, will trigger another build.

nerdypepper: works after re-adding.

PR #1115, opened 2023-11-06
Allow quick & shallow copies of Git repos to select which repos to index

rsdy:
Initiate the clone with:
```
; curl  -D - "localhost:7878/api/repos/sync?repo=github.com/bloopai/bloop&shallow=true"
```

This will create a shallow clone (1 commit depth), and a fast index of the repository in question, without looking at the body of any of the files.

After this process is done, the repository ends up in a `Shallow` state, which means it won't automatically be refreshed until the file filters are configured.

To configure file filters, use the following call:

```
; curl 'http://127.0.0.1:7878/api/repos/indexed?repo=github.com%2Fbloopai%2Fbloop' \
-X 'PATCH' \
-H 'Content-Type: application/json' \
-H 'Accept: application/json, text/plain, */*' \
--data-binary '{"file_filter": {"rules": [{"include_file": "flake.lock"}, {"include_regex": "^server/"}]}}'
```

The first call to this will automatically create a Git clone of 1000 commits in depth, and allows syncing the repo fully with the configured file filters.

Note that file filters are additive, and prioritise inclusion over exclusion. In other words, once something's been added to the index, can't be marked to be removed from the index without removing the repository completely!


anastasiya1155:
I tried it with your example query and also with the reverse one (exclude_file and exclude_regex) and as far as I can see it doesn't index anything by default and only files in 'include_regex' and 'include_file' were indexed. In the case of exclude_file and exclude_regex nothing was indexed at all. Can we change that so that everything is included by default and the user will use 'exclude_' fields to ignore some files and folders?

anastasiya1155 (approved):

server/bleep/src/repo.rs
```diff
@@ -217,6 +217,16 @@ pub struct Repository {
     #[serde(default)]
     pub file_filter: FileFilterConfig,
 
+    /// Indicate that this repository is to be cloned as a shallow copy
+    ///
+    /// Defaults to `false for existing repos.
```
nerdypepper, line 222:
```suggestion
    /// Defaults to `false` for existing repos.
```

server/bleep/src/indexes/file.rs
```diff
@@ -59,7 +59,13 @@ impl<'a> Workload<'a> {
...
                     .is_allowed(&self.relative_path)
-                    .map(|_| &b"__filter_override"[..])
+                    .map(|include| {
+                        if include {
+                            &b"__filter_override_include"[..]
+                        } else {
+                            &b"__filter_override_exclude"[..]
+                        }
+                    })
                     .unwrap_or(&b"__no_filter_override"[..]),
```
nerdypepper, line 69: `hash.update(b"foobar")` should work too, without the conversion of the array to `&[u8]`.

nerdypepper (approved):
code changes look good to me, will give this a test.

server/bleep/src/indexes/file.rs
nerdypepper: ah i suppose the `unwrap_or` will not typecheck without slices.

ggordonhall (commented):
Works well and code LGTM.

Before merging can you make sure that the `commits.rs` flow only runs after we've done a deep clone of the repo. As is it panics during the shallow clone step.

PR #1122, opened 2023-11-09
add repo sync analytics

nerdypepper:
tracks repo metadata on first index, or on full reindex (when the tantivy index hash is updated, for example), includes the following info:

- repo ref
- repo provider
- size in bytes
- number of qdrant chunks
- time taken to index

sample event:

    RepoEvent {
        name: "index",
        payload: [
            ("repo_ref", String("nerdypepper/dijo")),
            ("provider", String("github")),
            ("file_count", Number(35)),
            ("chunk_count", Number(236)),
            ("bytes", String("116.59KB")),
            ("sync_time", String("4.957807429s")),
        ],
    }

---

to test this changeset, set the following values in `local_config.json`:

    {
        "analytics_key": "..",
        "analytics_data_plane": ".."
    }

run rudderstack in "live event stream" mode, and index a repository.

### DIFF
--- a/server/bleep/src/cache.rs
+++ b/server/bleep/src/cache.rs
@@ -150,6 +150,19 @@ pub struct FileCache {
     embed_queue: EmbedQueue,
 }
 
+#[derive(Default)]
+pub struct InsertStats {
+    pub new: usize,
+    pub updated: usize,
+    pub deleted: usize,
+}
+
+impl InsertStats {
+    fn empty() -> Self {
+        Self::default()
+    }
+}
+
 impl<'a> FileCache {
     pub(crate) fn new(db: SqlDb, semantic: Semantic) -> Self {
         Self {
@@ -396,7 +409,7 @@ impl<'a> FileCache {
         buffer: &str,
         lang_str: &str,
         branches: &[String],
-    ) {
+    ) -> InsertStats {
         let chunk_cache = self.chunks_for_file(repo_ref, cache_keys).await;
         self.semantic
             .chunks_for_buffer(
@@ -420,10 +433,16 @@ impl<'a> FileCache {
                 info!(
                     repo_name,
                     relative_path, new, updated, deleted, "Successful commit"
-                )
+                );
+                InsertStats {
+                    new,
+                    updated,
+                    deleted,
+                }
             }
             Err(err) => {
-                warn!(repo_name, relative_path, ?err, "Failed to upsert vectors")
+                warn!(repo_name, relative_path, ?err, "Failed to upsert vectors");
+                InsertStats::empty()
             }
         }
     }
--- a/server/bleep/src/cache.rs
+++ b/server/bleep/src/cache.rs
@@ -429,16 +429,12 @@ impl<'a> FileCache {
             });
 
         match chunk_cache.commit().await {
-            Ok((new, updated, deleted)) => {
+            Ok(stats) => {
                 info!(
                     repo_name,
-                    relative_path, new, updated, deleted, "Successful commit"
+                    relative_path, stats.new, stats.updated, stats.deleted, "Successful commit"
                 );
-                InsertStats {
-                    new,
-                    updated,
-                    deleted,
-                }
+                stats
             }
             Err(err) => {
                 warn!(repo_name, relative_path, ?err, "Failed to upsert vectors");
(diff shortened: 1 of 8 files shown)


server/bleep/src/indexes/file.rs
```diff
@@ -173,14 +189,41 @@ impl Indexable for File {
...
+        analytics_event.add_payload("chunk_count", &repo_stats.chunks);
+        analytics_event.add_payload("bytes", &human_readable(repo_stats.size));
+        analytics_event.add_payload("index_job_kind", &repo_stats.index_job_kind());
+
+        let sync_time = start.elapsed();
+        info!(?repo.disk_path, "repo file indexing finished, took {:?}", sync_time);
+        analytics_event.add_payload("sync_time", &format!("{sync_time:?}"));
+
+        dbg!(&analytics_event);
+
```
nerdypepper, line 221:
```suggestion
```

server/bleep/src/indexes/file.rs
```diff
@@ -541,35 +584,83 @@ impl Indexer<File> {
...
+struct WorkerStats {
+    // size in bytes
+    size: usize,
+    // number of qdrant chunkc
+    chunks: usize,
+    // number of dir-entries reindexed by this worker
+    reindex_count: usize,
+    // set to true if this is the first index of this reporef
+    is_first_index: bool,
+
```
nerdypepper, line 597:
```suggestion
```

server/bleep/src/lib.rs
```diff
@@ -173,7 +174,10 @@ impl Application {
         config.source.save_index_version()?;
         debug!("index version saved");
 
-        let indexes = Indexes::new(&config, sql.clone()).await?.into();
+        let indexes = Indexes::new(&config, sql.clone())
+            .await?
+            .was_index_reset(dbg!(was_index_reset))
```
nerdypepper, line 179:
```suggestion
            .was_index_reset(was_index_reset)
```

rsdy: I think we should just move this into the constructor or create a separate constructor. We only ever initialize this once.

server/bleep/src/indexes/file.rs
```diff
@@ -791,29 +899,34 @@ impl RepoFile {
...
-            schema.last_commit_unix_seconds => last_commit,
-            schema.symbol_locations => bincode::serialize(&symbol_locations).unwrap(),
-            schema.symbols => symbols,
-            schema.branches => branches,
-            schema.is_directory => false,
-            schema.indexed => true,
+        stats.chunks += insert_stats.new;
+
+        Some((
+            stats,
```
rsdy, line 909: I think this may be cleaner if the workload had a `tx` reference, and instead of bubbling these back up, the event was generated and sent off at this point.

server/bleep/src/indexes/file.rs
```diff
@@ -173,14 +189,41 @@ impl Indexable for File {
...
+        info!(?repo.disk_path, "repo file indexing finished, took {:?}", sync_time);
+        analytics_event.add_payload("sync_time", &format!("{sync_time:?}"));
+
+        dbg!(&analytics_event);
+
+        // send an analytics event if this is the first time this repo is being indexed
+        if repo_stats.reindex_count > 0 {
+            let user = app.user().await;
+            app.with_analytics(|hub| hub.track_repo(analytics_event, &user));
+        }
```
rsdy, line 226: It may read nicer if we had an `indexes::analytics` (or something to that effect) module, which would contain this logic in a struct wrapped around the `rx` handle + all the other ancilliary structs. You could also measure index time between instantiation and commit as index time.

server/bleep/src/cache.rs
```diff
@@ -420,10 +433,16 @@ impl<'a> FileCache {
                 info!(
                     repo_name,
                     relative_path, new, updated, deleted, "Successful commit"
-                )
+                );
+                InsertStats {
```
rsdy, line 437: It may be worth turning the return value of `commit()` into `InsertStats`, so we don't have both the tuple and this.

rsdy (commented):
The fact that this patch is interwoven with the indexing logic so much makes it slightly difficult to track. I'd prefer containing this logic a bit more, and preferably separating it out to another module from `file` as much as we can.

server/bleep/src/indexes/file.rs
nerdypepper: agreed. have added a `tx` to each workload that `transmit_stats` when available.

rsdy (approved):
This looks a lot cleaner. Thanks

PR #1134, opened 2023-11-15
First steps for  multi-repo

rsdy:
This changeset is mostly backwards compatible, in particular it doesn't require a database migration. It should be enough to start testing and developing the logic for multi-repo searches by plumbing several API references through to the API layer.

The `/answer` endpoint breaks, and now expects `project: [repo_ref]` instead of `repo_ref` previously.


server/bleep/src/indexes/file.rs
```diff
@@ -213,9 +213,7 @@ impl Indexer<File> {
     /// If the regex filter fails to build, an empty list is returned.
     pub async fn fuzzy_path_match(
         &self,
-        repo_ref: &RepoRef,
         query_str: &str,
-        branch: Option<&str>,
```
calyptobai, line 218: Why remove these from `fuzzy_path_match`?

server/bleep/src/semantic.rs
```diff
@@ -412,9 +415,9 @@ impl Semantic {
         // In /answer we want to retrieve `limit` results exactly
         let results = self
             .search_with(
-                parsed_query,
```
calyptobai, line 415: Why do we strip the query here?

calyptobai (commented):
This generally LGTM

I haven't tested this locally, and have left some questions about 2 parts of the code that aren't quite clear to me

server/bleep/src/agent/prompts.rs
```diff
@@ -79,15 +81,20 @@ pub fn functions(add_proc: bool) -> serde_json::Value {
...
     let mut s = "".to_string();
 
     let mut paths = paths.into_iter().peekable();
 
     if paths.peek().is_some() {
-        s.push_str("## PATHS ##\nindex, path\n");
+        s.push_str("## PATHS ##\nindex, repo, path\n");
         for (i, path) in paths.enumerate() {
-            s.push_str(&format!("{}, {}\n", i, path));
+            let repo = path
```
ggordonhall, line 92: Let's just delete this logic and use the `display_name` of the `RepoRef`

server/bleep/src/agent/tools/code.rs
```diff
@@ -1,29 +1,42 @@
...
+        let repos = self.project.clone();
+
         let mut results = self
-            .semantic_search(query.into(), vec![], CODE_SEARCH_LIMIT, 0, 0.3, true)
+            .semantic_search(SemanticSearchParams {
+                query: Literal::from(&query.to_string()),
+                paths: vec![],
+                project: repos.clone(),
+                limit: CODE_SEARCH_LIMIT,
+                offset: 0,
```
ggordonhall, line 36: Could make `offset` and `retrieve_more` default params here.

server/bleep/src/agent/tools/proc.rs
```diff
@@ -30,21 +27,36 @@ impl Agent {
...
         .await?;
 
+        let relative_paths = paths.iter().map(|p| p.path.clone()).collect::<Vec<_>>();
+        // not sure we need this?
+        // let repos = paths.iter().map(|p| p.repo.clone()).collect::<Vec<_>>();
+
         let results = self
-            .semantic_search(query.into(), paths.clone(), 20, 0, 0.0, true)
+            .semantic_search(SemanticSearchParams {
+                query: Literal::from(&query.to_string()),
```
ggordonhall, line 36: We'll need to propagate `(repo, path)` pairs to the Qdrant query (to avoid the case where the same relative path is present in multiple repos). Maybe we could simplify this logic by having a `{repo}/{path}` field on each Qdrant payload that we can easily filter against.

server/bleep/src/indexes/file.rs
```diff
@@ -213,50 +214,46 @@ impl Indexer<File> {
     /// If the regex filter fails to build, an empty list is returned.
     pub async fn fuzzy_path_match(
         &self,
-        repo_ref: &RepoRef,
+        project: Project,
         query_str: &str,
-        branch: Option<&str>,
         limit: usize,
```
ggordonhall, line 219: We need to keep the ability to filter these searches by branch. We could keep branch as an `Option<&str>`, and filter searches for all paths on that branch across the provided `RepoRefs`.

ggordonhall (changes requested):

server/bleep/src/agent/tools/code.rs
rsdy: I wanted it easier to understand what the different parameters are. Unfortunately we can't `Default` this struct, so we need to pass these every time. To fix this we'd need to restructure the way this function works, or which component it belongs to.

server/bleep/src/indexes/file.rs
rsdy: We'd want to couple branches to `RepoRefs`, it doesn't make a lot of sense to, say, search across a list of repos in the `dev` branch, as they may not have one.

server/bleep/src/semantic.rs
rsdy: I've restored this.

PR #1136, opened 2023-11-16
Remove code snippets instead of redacting them in trimming

calyptobai:
Remove code chunks instead of redacting them. This should hopefully reduce the frequency at which we see `[REDACTED]` in model output.

Closes BLO-1842

### DIFF
--- a/server/bleep/src/agent/transcoder.rs
+++ b/server/bleep/src/agent/transcoder.rs
@@ -497,7 +497,8 @@ fn try_trim_code_xml(xml: &str) -> Result<String> {
     // We just remove code chunks completely.
 
     let xml = fixup_xml_code(xml);
-    let _code_chunk: CodeChunk = quick_xml::de::from_str(&xml).context("couldn't parse as XML code block")?;
+    let _code_chunk: CodeChunk =
+        quick_xml::de::from_str(&xml).context("couldn't parse as XML code block")?;
     Ok(String::new())
 }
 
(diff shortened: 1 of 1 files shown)


server/bleep/src/agent/transcoder.rs
```diff
@@ -494,45 +494,11 @@ fn xml_for_each(article: &str, f: impl Fn(&str) -> Option<String>) -> String {
...
-        CodeChunk::GeneratedCode { code: _, language } => {
-            format!(
-                "<GeneratedCode>\n\
-                <Code>[REDACTED]</Code>\n\
-                <Language>{language}</Language>\n\
-                </GeneratedCode>"
-            )
-        }
-    })
+    let xml = fixup_xml_code(xml);
```
ggordonhall, line 499: Do we need to do this work at all if we're returning an empty string?

ggordonhall (commented):
In testing this works well, but this is largely because the model rarely generates either `QuotedCode` or `GeneratedCode`: it prefers markdown code blocks so it's tricky to evaluate the effects of this change.

server/bleep/src/agent/transcoder.rs
calyptobai: We do need to do this for partially generated messages. E.g. if a message generates with `<GeneratedCode><Code>foo`, we need to call `fixup_xml_code` to complete the block so that it can still be parsed on the line right below. If parsing fails here (e.g. there is a *different* kind of XML block, perhaps HTML in the markdown), we know the block is not a code chunk and should be kept.

PR #1146, opened 2023-11-17
Automatically open bloop app after authentication with GitHub

anastasiya1155:
To improve the user experience when signing in with GitHub we now will re-open and focus bloop app once the user is authenticated on the FE. Tried using a couple of JS APIs from Tauri - all working unreliably or not working at all. The only solution that seems to work is focusing the window from the Rust side. For it to work there is now a new Tauri command that can be invoked from the FE when onboarding is closed.


rsdy (approved):
To be honest, I don't perfectly understand how this will work, but the PR looks coherent.

ggordonhall (approved):

PR #1149, opened 2023-11-21
Improve scrolling in chat and studio

anastasiya1155:
Use react-scroll-to-bottom to improve scrolling conversation to bottom.


stunningpixels:
Works, but package seems pretty stale?

anastasiya1155:
> Works, but package seems pretty stale?

Yes, it is indeed a bit stale, but works nevertheless, can rewrite it to our needs if it causes any issues

stunningpixels (approved):

ggordonhall (approved):
Works well!

PR #1154, opened 2023-11-27
Hybrid search (lexical + semantic) merged with RRF

rmuller-ml:
Hybrid search replaces semantic search with a mix between semantic search results and lexical search results Results are merged using reciprocal rank fusion (RRF).

To be able to merge results, each search (semantic and lexical) have to operate on the same documents. Here the documents are code snippets and to makes things easier, we use Qdrant as vector and snippet DB/index manager.

Wdrant lexical search does not rank lexical results based on known algorithms (tf-idf, bm25). So we rerank lexical results using word counts and number of query unique words matched.

This PR:

- Creates Qdrant snippet index (https://qdrant.tech/documentation/concepts/indexing/#full-text-index) using a word tokenizer
- creates a Qdrant lexical search function
- creates a rerank lexical results function
- creates a "merge semantic and lexical results with RRF" function
- changes search to perform hybrid search

### DIFF
--- a/server/bleep/src/semantic.rs
+++ b/server/bleep/src/semantic.rs
@@ -338,7 +338,7 @@ impl Semantic {
             must: build_conditions(parsed_query),
             ..Default::default()
         });
-        
+
         let response = self
             .qdrant
             .search_points(&SearchPoints {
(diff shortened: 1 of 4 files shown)


server/bleep/src/semantic.rs
```diff
@@ -424,7 +540,24 @@ impl Semantic {
...
+                    .collect::<Vec<_>>()
+            })?;
+        let results_lexical = Self::rank_lexical(results_lexical, &query);
+
+        let merged_results = Self::merge_rrf(results_lexical, dedup_results);
+
+        Ok(merged_results
+            .iter()
+            .take(limit.try_into().unwrap())
+            .cloned()
```
nerdypepper, line 559: can consume here instead of cloning?
```suggestion
            .into_iter()
            .take(limit.try_into().unwrap())
```

nerdypepper (approved):

ggordonhall (approved):

PR #1160, opened 2023-11-30
Update answer and studio to gpt4-turbo 24k

rmuller-ml:
- We separate the model from agent and answer to have different headrooms
- We don't use the full 128k context window (only 24k)

### DIFF
--- a/client/src/components/Chat/index.tsx
+++ b/client/src/components/Chat/index.tsx
@@ -95,7 +95,9 @@ const Chat = () => {
             }`
           : ''
       }&model=${
-        preferredAnswerSpeed === 'normal' ? 'gpt-4-turbo-24k' : 'gpt-3.5-turbo-finetuned'
+        preferredAnswerSpeed === 'normal'
+          ? 'gpt-4-turbo-24k'
+          : 'gpt-3.5-turbo-finetuned'
       }`;
       console.log(url);
       const eventSource = new EventSource(url);
--- a/client/src/components/Chat/index.tsx
+++ b/client/src/components/Chat/index.tsx
@@ -95,7 +95,7 @@ const Chat = () => {
             }`
           : ''
       }&model=${
-        preferredAnswerSpeed === 'normal' ? 'gpt-4' : 'gpt-3.5-turbo-finetuned'
+        preferredAnswerSpeed === 'normal' ? 'gpt-4-turbo-24k' : 'gpt-3.5-turbo-finetuned'
       }`;
       console.log(url);
       const eventSource = new EventSource(url);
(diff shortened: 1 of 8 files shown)


server/bleep/src/agent.rs
```diff
@@ -57,7 +57,8 @@ pub struct Agent {
     pub thread_id: uuid::Uuid,
     pub query_id: uuid::Uuid,
 
-    pub model: model::AnswerModel,
+    pub answer_model: model::LLMModel,
+    pub agent_model: model::LLMModel,
```
nerdypepper, line 61: i think its confusing that `Agent` has, within itself, an `agent_model` field, can we call this something else? such as `decsicion_model` or `reasoning_model`?

server/bleep/src/webserver/answer.rs
```diff
@@ -69,8 +69,10 @@ pub(super) async fn vote(
 pub struct Answer {
     pub q: String,
     pub repo_ref: RepoRef,
-    #[serde(default = "default_model")]
-    pub model: agent::model::AnswerModel,
+    #[serde(default = "default_answer_model")]
+    pub answer_model: agent::model::LLMModel,
+    #[serde(default = "default_agent_model")]
+    pub agent_model: agent::model::LLMModel,
```
nerdypepper, line 75: i believe this is a breaking change to the API. the FE currently this parameter in `model`, but the name has been changed to `answer_model`. here is what i think we should do:

* so we can either make the corresponding change on the FE here (cc @anastasiya1155)
* call it `answer_model` and use `#[serde(rename = "model")]`
* revert to old name   

side note: do we want to be able to configure `agent_model` with request parameter??

nerdypepper (changes requested):

client/src/components/Chat/index.tsx
```diff
@@ -94,8 +94,10 @@ const Chat = () => {
               queryIdToEdit ? `&parent_query_id=${queryIdToEdit}` : ''
             }`
           : ''
-      }&model=${
-        preferredAnswerSpeed === 'normal' ? 'gpt-4' : 'gpt-3.5-turbo-finetuned'
+      }&answer_model=${
```
rmuller-ml, line 97: Making it compatible with Answer struct interface change 'model' => 'answer_model'

server/bleep/src/webserver/answer.rs
rmuller-ml: I changed in the FE from model to answer_model (see comment in client/src/components/Chat/index.tsx)

server/bleep/src/agent.rs
rmuller-ml: I think `agent_model` is fine.
The abstraction is:  the user queries to get an `answer` and we have an `answer_model` used in the tool `answer`, to get the the correct context to answer the query we start an `agent` which has its own `agent_model`

Third opinion @ggordonhall ?

nerdypepper (approved):

PR #1161, opened 2023-12-06
Deduplicate after merging lexical results

ggordonhall:


ggordonhall:
> MMR over the merged results will affect the RRF (hybrid) ranking, because MMR it is basically ranking by cosine distance + diversity term so the lexical results will go to the bottom.
> 
> What exactly we are trying to add to the lexical results? Filter overlapping chunks? File path diversity? Prog. language diversity?

We're trying to ensure that we have `path` and `lang` diversity in the set of lexical results.

PR #1180, opened 2023-12-16
Additional chunks based on references and definitions in code and proc tools

rmuller-ml:
Some queries benefit from basic navigation (go to ref, go to def). Since adding a new tool to the agent makes the agent too complicated, we expand chunks retrieved by code search and proc with chunks containing definitions and references to symbols in the original chunks. To decide which symbol to add def/refs, we list all of the symbols in the original chunks, filter by some heuristics and use a basic LLM classifier to select a single symbol.

Pipeline
- code search retrieves 10 chunks
- we call `get_related_chunks` to get 3 additional chunks
  - we list all symbols from the original chunks in `extract_symbols`
  - we filter local symbols, symbols with no ref/defs and symbol with too many (>5) in `extract_symbols`
  - we use a llm classifier in `filter_stmbols` to select a symbol to expand
  - we expand the symbol into snippets in `expand_symbol_into_chunks`
 
- same things is done in proc
 
Implementation

`agent/symbol.rs` is the main file with the changes
hoverable endpoint is used to collect all symbols from a file
token-info endpoint is used to get refs/defs for a given symbol

  Remaining things to do
  
- [x]  revert path filter in proc back to text search (instead of keyword search)
- [x] handle errors by cleaning the code from unwraps 
  - [x] llm call fails
  - [x] function call json fails
  - [x] selected symbol alias not in the set of symbols
  - [x] list of symbols is empty
- [x ] make it faster (load documents a single time per chunk)

<a href="https://gitpod.io/#https://github.com/BloopAI/bloop/pull/1180"></a>


server/bleep/src/agent/symbol.rs
```diff
@@ -0,0 +1,361 @@
...
+            .iter()
+            .filter(|range| {
+                (range.start.byte >= chunk.start_byte) && (range.start.byte < chunk.end_byte)
+            })
+            .map(|range| {
+                token_info(
+                    TokenInfoRequest {
+                        relative_path: chunk.path.clone(),
+                        repo_ref: repo_ref.clone(),
+                        branch: None,
```
rmuller-ml, line 46: get the branch

server/bleep/src/agent/symbol.rs
```diff
@@ -0,0 +1,370 @@
...
+                    )
+                }),
+        )
+        .await;
+
+        // filter references and definitions
+        let mut symbols = related_symbols
+            .into_iter()
+            .filter_map(Result::ok)
+            .zip(hoverable_ranges.into_iter().filter(|range| {
```
ggordonhall, line 73: Can you comment explaining these filters?

server/bleep/src/agent/symbol.rs
```diff
@@ -0,0 +1,370 @@
...
+
+        debug!("Attatched {} symbols", symbols.len());
+
+        Ok(ChunkWithSymbols {
+            chunk: chunk.clone(),
+            symbols,
+        })
+    }
+
+    pub async fn expand_symbol_into_chunks(&self, ref_def_metadata: Symbol) -> Vec<CodeChunk> {
```
ggordonhall, line 106: @nerdypepper Is there some logic that we can reuse for this?

server/bleep/src/agent/symbol.rs
```diff
@@ -0,0 +1,370 @@
...
+                    c.path.clone(),
+                    c.snippet.clone(),
+                    symbols_string
+                )
+            })
+            .collect::<Vec<_>>()
+            .join("\n\n");
+
+        // instruction
+        let prompt = format!("Snippets:\n\n{}\n\nInstruction: Above there are some code chunks and some symbols extracted from the chunks. Your job is to select the most relevant symbol to the user query. Do not answer with the siymbol name, use the symbol key/alias.\n\nQuery:{}", chunks_string.as_str(),  query);
```
ggordonhall, line 206: Let's move this to `prompts.rs`

server/bleep/src/agent/symbol.rs
```diff
@@ -0,0 +1,370 @@
...
+            .iter()
+            .take(MAX_CHUNKS)
+            .map(|c| {
+                let chunk = CodeChunk {
+                    path: c.path.clone(),
+                    alias: self.get_path_alias(c.path.as_str()),
+                    snippet: c.snippet.clone(),
+                    start_line: c.start_line,
+                    end_line: c.end_line,
+                    start_byte: 0,
```
ggordonhall, line 304: Can we return the correct byte range here?

server/bleep/src/webserver/intelligence.rs
```diff
@@ -19,25 +19,25 @@ use serde::{Deserialize, Serialize};
...
-    end: usize,
+    pub start: usize,
+    pub end: usize,
 }
 
 /// The response from the `local-intel` endpoint.
 #[derive(Serialize, Debug)]
-pub(super) struct TokenInfoResponse {
-    data: Vec<FileSymbols>,
+pub struct TokenInfoResponse {
```
ggordonhall, line 39: I think we can revert this change

server/bleep/src/webserver/answer.rs
```diff
@@ -451,6 +451,8 @@ pub async fn explain(
         start_line: params.line_start,
         end_line: params.line_end,
         snippet,
+        start_byte: 0,
```
ggordonhall, line 454: Again, can we set the correct byte ranges here? Are they ever used?

ggordonhall (changes requested):
Can you add a detailed description of the approach in a comment in `symbol.rs`?

server/bleep/src/agent/symbol.rs
```diff
@@ -0,0 +1,370 @@
...
+
+        // get expanded chunks for selected symbol
+        let extra_chunks = self.expand_symbol_into_chunks(selected_symbol).await;
+
+        // take 3 chunks, update path aliases, update enchange chunks
+        let extra_chunks = extra_chunks
+            .iter()
+            .take(MAX_CHUNKS)
+            .map(|c| {
+                let chunk = CodeChunk {
```
nerdypepper, line 298: if we update the path alias in the `expand_symbols_into_chunks` function; we will have to simply clone the `chunk` here. if can cannot update the path alias earlier, we can use this shorthand notation:

```rust
let chunk = CodeChunk {
    alias: self.get_path_alias(..),
    ..c       // the rest of the values are cloned from `c`
}

File: server/bleep/src/agent/tools/code.rs [hunk 9]
```diff
@@ -71,9 +73,13 @@ impl Agent {
                 .push(chunk.clone())
         }
 
+        let extra_chunks = self.get_related_chunks(chunks.clone()).await;
+
+        chunks.extend(extra_chunks);
+
         let response = chunks
-            .iter()
-            .filter(|c| !c.is_empty())
```
nerdypepper, line 76: can we keep the `filter` call?

server/bleep/src/webserver/intelligence.rs
```diff
@@ -335,14 +307,52 @@ pub(super) async fn token_value(
...
+    params: TokenInfoRequest,
+    repo_ref: &RepoRef,
+    indexes: Arc<Indexes>,
+    source_doc: &ContentDocument,
+    all_docs: &Vec<ContentDocument>,
+) -> anyhow::Result<Vec<FileSymbols>> {
+    let source_document_idx = all_docs
+        .iter()
+        .position(|doc| doc.relative_path == source_doc.relative_path)
+        .unwrap(); // FIXME: handle this
```
nerdypepper, line 320: can we preserve the orignal error handling code here? `.ok_or(anyhow::anyhow!("invalid language"))`?

server/bleep/src/agent/symbol.rs
rmuller-ml: I added

```
        // filter references and definitions
        // 1: symbol shouldn't be in the same file
        // 2: number of refs/defs should be less than 5 to avoid very common symbols (iter, unwrap...)
        // 3: also filter out symbols without refs/defs
```

server/bleep/src/agent/symbol.rs
rmuller-ml: I changed token-info so we can define `context_before` and `context_after`

server/bleep/src/agent/symbol.rs
```diff
@@ -0,0 +1,370 @@
...
+                    .data
+                    .iter()
+                    .map(|occurrence| {
+                        let chunk_content = content[occurrence.range.start.line
+                            ..(occurrence.range.end.line + NUMBER_CHUNK_LINES).min(n_lines)]
+                            .to_vec()
+                            .join("\n");
+                        CodeChunk {
+                            path: filename.clone(),
+                            alias: 0,
```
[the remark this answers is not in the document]
rmuller-ml: We cannot set here because we will later truncate the chunks, so we want to avoid adding the alias/path in the prompt if the agent cannot see the chunk

server/bleep/src/agent/symbol.rs
rmuller-ml: done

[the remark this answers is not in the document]
rmuller-ml: done

server/bleep/src/agent/symbol.rs
```diff
@@ -0,0 +1,370 @@
...
+        let filter_function = serde_json::from_value::<Vec<llm_gateway::api::Function>>(serde_json::json!([
+            {
+                "name": "filter",
+                "description":  "Select the symbol most likely to contain information to answer the query",
+                "parameters": {
+                    "type": "object",
+                    "properties": {
+                        "symbol": {
+                            "type": "integer",
+                            "description": "The chunk alias"
```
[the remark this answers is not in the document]
rmuller-ml: i changec from `chunk` to `symbol`, thanks

server/bleep/src/agent/symbol.rs
rmuller-ml: thanks, much cleaner now

server/bleep/src/agent/tools/code.rs
rmuller-ml: done

server/bleep/src/webserver/intelligence.rs
rmuller-ml: done

server/bleep/src/agent/symbol.rs
```diff
@@ -0,0 +1,370 @@
...
+            hoverable_ranges
+                .iter()
+                .filter(|range| {
+                    (range.start.byte >= chunk.start_byte) && (range.start.byte < chunk.end_byte)
+                })
+                .map(|range| {
+                    get_token_info(
+                        TokenInfoRequest {
+                            relative_path: chunk.path.clone(),
+                            repo_ref: String::new(), // FIXME
```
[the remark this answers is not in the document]
rmuller-ml: done

server/bleep/src/webserver/intelligence.rs
rmuller-ml: done

server/bleep/src/webserver/answer.rs
rmuller-ml: the only place those are used is when getting (filtering) the subset of symbols belonging to a chunk out of all symbols in a file in `tools/symbol.rs`, that is why I added the field. 

I could change the type to `Option<usize>` for the cases where we want to create a chunk, but we don't know the exact start end bytes.

server/bleep/src/webserver/intelligence.rs
```diff
@@ -335,14 +315,60 @@ pub(super) async fn token_value(
...
+            start_byte: params.start,
+            end_byte: params.end,
+        },
+        all_docs,
+        source_document_idx,
+        snipper,
+    };
+
+    let data = ctx.token_info();
+    if data.is_empty() {
```
rmuller-ml, line 347: @nerdypepper 

`data` here is empty, so we are using search_nav (tantivy search). When `token_info` returns a non empty vector? should it be faster than `search_nav`?

ggordonhall (approved):

PR #1186, opened 2023-12-19
rework sync logic for docs

nerdypepper:
- replace `/sync` with `/enqueue`; a non-streaming replacement to add items to the doc-sync queue
- introduce `/status` and `/cancel`; to stream updates for a syncing document or to cancel a sync job
- convert `/resync` to http from sse
- internal updates to `/list` to work with the new queue system

--- 

i have dropped `TODO` comments in a few locations that require error handling; i will add these in a follow up PR:
- if a sync job fails (e.g.: zero docs were scraped); the join handle should be deleted and the progress should report that
- if a progress updated failed for some reason, we should be able to know so in sentry 

<a href="https://gitpod.io/#https://github.com/BloopAI/bloop/pull/1186"></a>

### DIFF
--- a/server/bleep/src/indexes/doc.rs
+++ b/server/bleep/src/indexes/doc.rs
@@ -8,7 +8,7 @@ use tantivy::{
     tokenizer::NgramTokenizer,
 };
 use thiserror::Error;
-use tokio::sync::Mutex;
+use tokio::sync::{Mutex, RwLock};
 use tracing::{error, info, trace};
 
 use crate::{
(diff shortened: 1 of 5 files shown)


server/bleep/src/indexes/doc.rs
```diff
@@ -248,94 +277,191 @@ impl Doc {
...
+        }
+        .execute(&mut transaction)
+        .await?
+        .last_insert_rowid();
+
+        let (tx, rx) = tokio::sync::watch::channel(Progress::Init(id));
+        let self_ = self.clone();
+        let url_ = url.clone();
+        let join_handle = tokio::task::spawn(async move {
+            self_.begin_sync_job(id, url_, tx, transaction).await; // TODO: handle err
```
ggordonhall, line 298: Let's handle this

ggordonhall (approved):
Looks good to me. Can you address your TODOs?

PR #1204, opened 2024-01-17
Fix code search on local repos

calyptobai:
We were making semantic queries with the full stringified repo ref. Instead, we should have been constructing a semantic query using the repository display name.

It seems that this was fixed coincidentally in #1190 via a condition (which might now be possible to remove).

### DIFF
--- a/server/bleep/src/agent.rs
+++ b/server/bleep/src/agent.rs
@@ -432,7 +432,7 @@ impl Agent {
             target: Some(query),
             repos: repos
                 .iter()
-                .map(RepoRef::display_name)
+                .map(RepoRef::indexed_name)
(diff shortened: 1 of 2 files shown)


server/bleep/src/agent.rs
```diff
@@ -432,7 +432,7 @@ impl Agent {
             target: Some(query),
             repos: repos
                 .iter()
-                .map(RepoRef::to_string)
+                .map(RepoRef::display_name)
```
ggordonhall, line 435:
```suggestion
                .map(RepoRef::indexed_name)
```

We can use `indexed_name` and remove this block: https://github.com/BloopAI/bloop/blob/projects-fix-local-code-search/server/bleep/src/semantic.rs#L811-L817

PR #1205, opened 2024-01-18
Add fields to list studios (projects)

anastasiya1155:
Return token counter, doc_context, and full context files in list studios route

<a href="https://gitpod.io/#https://github.com/BloopAI/bloop/pull/1205"></a>


calyptobai (approved):
Looks good, we need to run `cargo fmt` as a final commit before merging this.

PR #1275, opened 2024-06-11
Oss fixes

anastasiya1155:
hardcode user_id to 1


ggordonhall (approved):

ycclnn:
May I ask why this commit is reverted

anastasiya1155:
> May I ask why this commit is reverted

It was an incorrect solution, userId should be taken from GitHub using the token in config and not hardcoded

